PR #132, opened 2024-07-22
chore: docker-compose - increase nginx client_max_body_size

mfiedorowicz:
Increase nginx's `client_max_body_size` from default `1m` to `25m`


leoparente (approved):
LGTM

PR #135, opened 2024-07-22
feat: add tests to netbox wrappers

leoparente:


leoparente:
Closing it for now, as it was added as internal package. We should aim to do that as package netbox_test and cover each netbox structure

PR #140, opened 2024-07-24
feat: diode-proto - set explicit fields presence

mfiedorowicz:
This is to track protobuf fields / detect if field has value (provided vs default zero-value), which will allow us to reset fields in NetBox models if instructed by providing zero-values, i.e. an interface with `enabled` field with `false` value to disable interface (currently does nothing it interface is enabled), similarly with any other optional field (an interface's `description` with `Lorem ipsum` cannot be reset to an empty string, while these changes will allow to provide `description=""`  to reset in the NetBox).

ref: https://protobuf.dev/programming-guides/field_presence/#presence-in-proto3-apis

Bonus: `make` recipes for development docker compose (building image instead of using `latest` from container registry) and don't build image with `make docker-compose-up` (pulls `latest` tags from container registry).

Test Execution Report


leoparente (approved):
LGTM. Nice adding on compose for developers

PR #143, opened 2024-07-26
feat: add unit tests for ingester component

leoparente:


diode-server/ingester/component_test.go
```diff
@@ -0,0 +1,338 @@
...
+	if err != nil {
+		return strconv.Itoa(0), err
+	}
+
+	addr := listener.Addr().(*net.TCPAddr)
+
+	err = listener.Close()
+	if err != nil {
+		return strconv.Itoa(0), err
+	}
```
mfiedorowicz: minor code style:
```suggestion
	if err = listener.Close(); err != nil {
		return strconv.Itoa(0), err
	}
```

mfiedorowicz (approved):

leoparente:
PS - I was not able to force/test `newAuthUnaryInterceptor` from public methods.

PR #145, opened 2024-08-02
feat: docker compose - remove redundant configurediodeplugin command call

mfiedorowicz:
Reflecting [changes](https://github.com/netboxlabs/diode-netbox-plugin/pull/17) introduced in the diode-netbox-plugin


leoparente (approved):
LGTM 🧹

PR #146, opened 2024-08-02
🚚 release: protobuf explicit field presence

mfiedorowicz:
- https://github.com/netboxlabs/diode/pull/140 - added explicit field presence to various fields in supported object type envelopes, solving issue when ingesting data with explicitly provided default values for scalar types (`false` for boolean, `""` for string, etc.) that were ignored, i.e. when attempting to disable existing enabled interface or _removing_ existing description


leoparente (approved):
Lgtm

PR #149, opened 2024-08-05
fix: docker - configure redis append only file directory

mfiedorowicz:
Setting append only file directory to `/data` to match mounted volume so data is loaded between docker compose down and up.

Also setting snapshotting to 60s for at least 1 write operation.


leoparente (approved):
LGTM

PR #150, opened 2024-08-07
feat: Add new schema data to ingest-entity

leoparente:
Also add redis commander to docker compose to make debug easier

### DIFF
--- a/diode-server/docker/docker-compose.yaml
+++ b/diode-server/docker/docker-compose.yaml
@@ -76,7 +76,26 @@ services:
     image: redis/redis-stack-server:latest
     links:
       - diode-redis
-    command: redis-cli -h "$REDIS_HOST" -p "$REDIS_PORT" -a "$REDIS_PASSWORD" FT.CREATE ingest-entity ON JSON PREFIX 1 "ingest-entity:" SCHEMA $$.data_type AS data_type TEXT $$.state AS state NUMERIC
+    entrypoint:
+      - sh
+      - -c
+      - |
+        redis-cli -h "$REDIS_HOST" -p "$REDIS_PORT" -a "$REDIS_PASSWORD" <<EOF
+        FT.CREATE ingest-entity ON JSON PREFIX 1 "ingest-entity:" SCHEMA \
+        $.data_type AS data_type TEXT \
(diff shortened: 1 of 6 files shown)


diode-server/reconciler/ingestion_processor_test.go
```diff
@@ -181,7 +181,7 @@ func TestIngestionProcessorStart(t *testing.T) {
 	// Add a message to the Redis stream
 	metadata := []string{
 		"request", string(reqBytes),
-		"ingestion_ts", "test-timestamp",
+		"ingestion_ts", 1720425600,
```
leoparente:
```suggestion
		"ingestion_ts", "1720425600",
```

diode-server/docker/docker-compose.yaml
```diff
@@ -76,7 +76,26 @@ services:
...
+        EOF
+  redis-commander:
+    container_name: redis-commander
+    hostname: redis-commander
+    image: ghcr.io/joeferner/redis-commander:latest
+    restart: always
+    environment:
+    - REDIS_HOSTS=local:$REDIS_HOST:$REDIS_PORT:0:$REDIS_PASSWORD
+    ports:
+    - "8082:8081"
```
mfiedorowicz: `8081` should be matching `$REDIS_PORT` from the env var.

mfiedorowicz: As redis-commander should be available for development only instead of official docker-compose, I'd suggest to add `docker-compose.dev.yaml` as additional service and extend default `docker-compose.yaml` in following Makefile recipes:
- `docker-compose-dev-up`
- `docker-compose-dev-down`

leoparente: why so? don't get it.. it is the internal port for the webserver. the redis port is passed here:

` REDIS_HOSTS=local:$REDIS_HOST:$REDIS_PORT:0:$REDIS_PASSWORD`

mfiedorowicz: Ah, conflated redis-commander port with redis one, sorry!

leoparente: done!

mfiedorowicz (approved):

PR #154, opened 2024-08-13
feat: add RetrieveIngestionLogs to reconciler proto

leoparente:


diode-server/reconciler/server.go
```diff
@@ -109,6 +109,12 @@ func (s *Server) RetrieveIngestionDataSources(_ context.Context, in *reconcilerp
 	return &reconcilerpb.RetrieveIngestionDataSourcesResponse{IngestionDataSources: dataSources}, nil
 }
 
+// RetrieveIngestionLogs retrieves logs
+func (s *Server) RetrieveIngestionLogs(_ context.Context, _ *reconcilerpb.RetrieveIngestionLogsRequest) (*reconcilerpb.RetrieveIngestionLogsResponse, error) {
+	logs := make([]*reconcilerpb.IngestionLog, 0)
```
leoparente, line 141: @mfiedorowicz do we need API key for logs as well?

mfiedorowicz: Yes, `NETBOX_TO_DIODE` API key please

diode-server/proto/reconciler/v1/reconciler.proto
```diff
@@ -36,8 +36,38 @@ message RetrieveIngestionDataSourcesResponse {
   repeated IngestionDataSource ingestion_data_sources = 1;
 }
 
+// An ingestion log
+message IngestionLog {
+  string id = 1; // Unique identifier for the log
```
mfiedorowicz: What does it correspond to in redis? Is it already or would be store somewhere?

diode-server/proto/reconciler/v1/reconciler.proto
```diff
@@ -36,8 +36,38 @@ message RetrieveIngestionDataSourcesResponse {
...
+  string state = 2; // State of the log
```
mfiedorowicz: Could we add validator or go even further and instead of string define enum corresponding to ingestion state constants used in the processor https://github.com/netboxlabs/diode/blob/develop/diode-server/reconciler/ingestion_processor.go#L36-L47

diode-server/proto/reconciler/v1/reconciler.proto
```diff
@@ -36,8 +36,38 @@ message RetrieveIngestionDataSourcesResponse {
...
+  string data_type = 3; // Data type of the log
+  string request_id = 4; // Request ID associated with the log
+  int64 ingestion_ts = 5; // Timestamp of the ingestion
+  string value = 6; // The stored value in the JSON key
```
mfiedorowicz: Currently, we store original ingested entity and changeset ID after successful reconciliation but planning to store generated changeset payload as well (separately to this PR). I'd suggest to rename `value` (as too generic) with `entity` of existing proto `Entity` type to be more specific, and later on we can add `changeset` field once we have more data to work on.

diode-server/proto/reconciler/v1/reconciler.proto
```diff
@@ -36,8 +36,38 @@ message RetrieveIngestionDataSourcesResponse {
...
+}
+
+// The request to retrieve ingestion logs
+message RetrieveIngestionLogsRequest {
+  int32 number_of_entities = 1; // Override the default page size
```
mfiedorowicz:
```suggestion
  int32 page_size = 1; // Override the default page size
```

Assuming default page size to be defined in the SDK/client and if not present there, we'd also set default page size on the server side and the `page_size` in the request to allow user to override SDK default.

diode-server/proto/reconciler/v1/reconciler.proto
leoparente: you mean to generate a dependency between reconciler and ingester? currently we do not have import ingester in reconciler proto

leoparente: done

diode-server/proto/reconciler/v1/reconciler.proto
leoparente: It was invalid id. I've removed it

diode-server/proto/reconciler/v1/reconciler.proto
leoparente: done

diode-server/proto/reconciler/v1/reconciler.proto
```diff
@@ -36,8 +37,67 @@ message RetrieveIngestionDataSourcesResponse {
...
+      string change_id = 2;
+    }
+    repeated Error errors = 3;
+  }
+  string message = 1;
+  int32 code = 2;
+  Details details = 3;
+}
+
+enum State {
```
leoparente: @mfiedorowicz what do you think about this?

the proto generated code is:


```
type State int32

const (
	State_NEW        State = 0
	State_RECONCILED State = 1
	State_FAILED     State = 2
	State_NO_CHANGES State = 3
)

// Enum value maps for State.
var (
	State_name = map[int32]string{
		0: "NEW",
		1: "RECONCILED",
		2: "FAILED",
		3: "NO_CHANGES",
	}
	State_value = map[string]int32{
		"NEW":        0,
		"RECONCILED": 1,
		"FAILED":     2,
		"NO_CHANGES": 3,
	}
)

func (x State) Enum() *State {
	p := new(State)
	*p = x
	return p
}
```

diode-server/proto/reconciler/v1/reconciler.proto
```diff
@@ -36,8 +37,67 @@ message RetrieveIngestionDataSourcesResponse {
   repeated IngestionDataSource ingestion_data_sources = 1;
 }
 
+// ChangeSetError represents an error when applying a change set
+message ChangeSetError {
+  message Details {
```
leoparente: thoughts about nested structures vs global? I've added as nested because it does not make that sense being used separetely

mfiedorowicz: Yes, it should be fine 👍

diode-server/proto/reconciler/v1/reconciler.proto
mfiedorowicz: Looks good. There is general recommendation to add `*_UNSPECIFIED = 0` first enum value (https://protobuf.dev/programming-guides/dos-donts/#unspecified-enum), but in our case it's for output (ingestion log) rather than input, so it's fine IMHO.

leoparente: in our case I don't think it makes sense and also when filtering it is optional value 0 is NEW for us

diode-proto/diode/v1/reconciler.proto
```diff
@@ -2,6 +2,7 @@ syntax = "proto3";
 
 package diode.v1;
 
+import "diode/v1/ingester.proto";
 import "validate/validate.proto";
 
 option go_package = "github.com/netboxlabs/diode/diode-server/reconciler/v1/reconcilerpb";
```
mfiedorowicz: 🙏 (forgot to change it previous PR)

```suggestion
option go_package = "github.com/netboxlabs/diode/diode-server/gen/diode/v1/reconcilerpb";
```

mfiedorowicz (approved):
Looks good 👏

PR #158, opened 2024-08-29
feat: reorg proto structure

mfiedorowicz:
- upgrade buf to v2 ([details](https://buf.build/docs/migration-guides/migrate-v2-config-files))
- reorganize proto definitions for ingester and reconciler services under `diode.v1` package
- use ingester go code from `netboxlabs/diode-sdk-go` (same existing `diodepb` go package) removing same code being generated into go sdk and diode-server internally
- reconciler.proto generated into `diode-server/reconciler/v1/reconcilerpb` (as-is)


diode-server/ingester/component.go
```diff
@@ -16,7 +16,7 @@ import (
 	"google.golang.org/grpc/reflection"
 	"google.golang.org/protobuf/proto"
 
-	"github.com/netboxlabs/diode/diode-server/gen/diode/v1/diodepb"
+	"github.com/netboxlabs/diode-sdk-go/diode/v1/diodepb"
```
leoparente: So the idea is that there will be no internal gen for diode-proto. All the code will exist in go sdk, is that right?

diode-server/Makefile
```diff
@@ -34,8 +34,8 @@ test:
 
 test-coverage:
 	@mkdir -p .coverage
-	@go test -race -cover -json -coverprofile=.coverage/cover.out.tmp ./... | grep -Ev "diodepb|reconcilerpb|cmd|mocks|sentry" | tparse -format=markdown > .coverage/test-report.md
-	@cat .coverage/cover.out.tmp | grep -Ev "diodepb|reconcilerpb|cmd|mocks|sentry" > .coverage/cover.out
+	@go test -race -cover -json -coverprofile=.coverage/cover.out.tmp ./... | grep -Ev "reconcilerpb|cmd|mocks|sentry" | tparse -format=markdown > .coverage/test-report.md
```
leoparente: The following code should be removed from this file
```

reconciler-proto-gen:
	@cd proto/ && buf format -w && buf generate --template buf.gen.yaml
```

diode-server/ingester/component.go
mfiedorowicz: To be precise: there will be no internal gen for ingester proto, only reconciler proto being generated into `diode-server/reconciler/v1/reconcilerpb`. Considered generating everything into diode-sdk-go as it's one `diode.v1` proto package, but somehow having reconciler there doesn't fit well, as it's more internal service. Thoughts?

leoparente: I think we should do either:
- generate everything in go-sdk and use the dependency
- generate both internally and does not depended on SDK  (and sdk generates only ingester)

mfiedorowicz: Variant 2 it is, thanks!

diode-server/proto/reconciler/v1/reconciler.proto
```diff
@@ -29,6 +30,7 @@ message RetrieveIngestionDataSourcesRequest {
     max_len: 255
   }];
   string sdk_version = 3 [(validate.rules).string = {pattern: "^(\\d)+\\.(\\d)+\\.(\\d)+$"}];
+  diode.v1.Entity entity = 4;
```
leoparente: should we remove it?

mfiedorowicz: yes! good spot, thanks

mfiedorowicz: actually there also shouldn't be diode-server/proto dir anymore

leoparente (approved):

PR #162, opened 2024-09-03
feat: add integration test for Virtual Machine Objects

leoparente:
image

### DIFF
--- a/diode-server/reconciler/changeset/changeset_virt_test.go
+++ b/diode-server/reconciler/changeset/changeset_virt_test.go
@@ -464,6 +464,439 @@ func TestVirtualizationPrepare(t *testing.T) {
 			wantErr: false,
 		},
 		{
+			name: "[P4] ingest virtualization.virtualmachine with name and cluster - existing objects not found - create",
+			rawIngestEntity: []byte(`{
+				"request_id": "cfa0f129-125c-440d-9e41-e87583cd7d89",
+				"data_type": "virtualization.virtualmachine",
+				"entity": {
+					"VirtualMachine": {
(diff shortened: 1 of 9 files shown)


diode-server/reconciler/changeset/changeset_virt_test.go
mfiedorowicz, line 1: As we don't create cluster placeholder for virtual machine (if not cluster not ingested alongside VM) anymore, would be advisable to have tests for cases:
- ingest virtualization.virtualmachine with name only (without cluster): this is already covered with latest changes here ✅ 
- ingest virtualization.virtualmachine with name and cluster (both don't exist): to ensure we generate changeset to create both cluster and VM
- ingest virtualization.virtualmachine with name and cluster (cluster exists, VM doesn't): changeset should contain creation of the VM with an existing cluster object
- ingest virtualization.virtualmachine with name and cluster (cluster doesn't exist, VM exists): changeset should contain creation of the cluster and update of the VM with a new cluster object

Does it sound feasible?

mfiedorowicz (commented):
Thanks for all of these tests 🙌 Left one question/suggestion.

diode-server/reconciler/changeset/changeset_virt_test.go
leoparente: I've added the tests, but it raised a question

ingest virtualization.virtualmachine with name and cluster (cluster doesn't exist, VM exists): changeset should contain creation of the cluster and update of the VM with a new cluster object
 - is expected to update for VM be performed?

mfiedorowicz (approved):
🙌

PR #166, opened 2024-09-06
feat: implement reconciler retrieve ingestion logs

leoparente:


leoparente:
Still missing parse entity

mfiedorowicz (approved):

diode-server/reconciler/logs_retriever.go
```diff
@@ -0,0 +1,169 @@
...
+)
+
+type extraAttributesWrapper struct {
+	ExtraAttributes string `json:"$"`
+	IngestionTs     string `json:"ingestion_ts"`
+}
+type redisLogResult struct {
+	ExtraAttributes extraAttributesWrapper `json:"extra_attributes"`
+	ID              string                 `json:"id"`
+	Values          []interface{}          `json:"values"`
```
mfiedorowicz: Do we need this at all?

leoparente: I don't think so, I will remove it as well. I've already removed some unused fields

mfiedorowicz: Probably we can remove from test inputs as well

leoparente: I kept on log to mock the full data from the redis FT SEARCH request

PR #169, opened 2024-09-13
feat: add ingestion logs metrics

leoparente:
- Add `IngestionMetrics` structure to reconciler proto
- Add `only_metrics` filter to IngestionLogsRequest to trigger retrieval of all metrics without any entity
- Add `metrics` return to `IngestionLogsResponse`
- Allow user to pass PageSize '0' by change this field to optional. 0 Allows returning the total records without returning any specific entity
If a Request has state filter, the proper stat will be returned for it.


diode-server/reconciler/logs_retriever.go
```diff
@@ -13,6 +13,7 @@ import (
 	"google.golang.org/protobuf/encoding/protojson"
 
 	"github.com/netboxlabs/diode/diode-server/gen/diode/v1/reconcilerpb"
+	"github.com/redis/go-redis/v9"
```
mfiedorowicz: nit: goimports

leoparente: i've run `goimports -l -w `, no changes happened..

mfiedorowicz: Let's treat `github.com/netboxlabs/diode` as local, i.e. with `goimports -l -local "github.com/netboxlabs/diode" -w`.

The expected imports groups:
```
stdlib

3rd party libs

local libs
```

diode-proto/diode/v1/reconciler.proto
```diff
@@ -61,6 +61,15 @@ enum State {
   NO_CHANGES = 4;
 }
 
+// Log Stats
+message Stats {
```
mfiedorowicz: Shall we be more explicit with naming, e.g. `IngestionStats` or even `IngestionMetrics`, thoughts?

diode-proto/diode/v1/reconciler.proto
```diff
@@ -61,6 +61,15 @@ enum State {
...
+  optional int32 total = 1;
+  optional int32 new = 2;
+  optional int32 reconciled = 3;
+  optional int32 failed = 4;
+  optional int32 no_changes = 5;
```
mfiedorowicz: What's reason behind use of `optional` here? Without it we would default to 0 which is fine.

diode-proto/diode/v1/reconciler.proto
```diff
@@ -77,22 +86,21 @@ message IngestionLog {
 
 // The request to retrieve ingestion logs
 message RetrieveIngestionLogsRequest {
-  int32 page_size = 1 [(validate.rules).int32 = {
-    gte: 1
-    lte: 1000
-  }]; // Number of logs per page, default is 100
+  optional int32 page_size = 1; // Number of logs per page, default is 100
```
mfiedorowicz, line 89: 👍

diode-proto/diode/v1/reconciler.proto
```diff
@@ -77,22 +86,21 @@ message IngestionLog {
...
   optional State state = 2; // Optional filter by state field
   string data_type = 3; // Optional filter by data type field
   string request_id = 4; // Optional filter by request ID
   int64 ingestion_ts_start = 5; // Optional start of ingestion timestamp range
   int64 ingestion_ts_end = 6; // Optional end of ingestion timestamp range
   string page_token = 7; // Token to fetch the next page of results
+  bool summary = 8; // Optional flag to return all stats
```
mfiedorowicz: s/summary/include_metrics ?

diode-proto/diode/v1/reconciler.proto
```diff
@@ -77,22 +86,21 @@ message IngestionLog {
...
 }
 
 // The response from the retrieve ingestion logs request
 message RetrieveIngestionLogsResponse {
   repeated IngestionLog logs = 1; // List of ingestion logs
-  string next_page_token = 2; // Token for the next page of results, if any
+  Stats stats = 2; // Stats for the logs
```
mfiedorowicz:
```suggestion
  IngestionMetrics metrics = 2; // Ingestion metrics
```

diode-proto/diode/v1/reconciler.proto
mfiedorowicz: Would be great to have similar corresponding values as in https://github.com/netboxlabs/diode/blob/develop/diode-proto/diode/v1/reconciler.proto#L56-L62

```
enum State {
  UNSPECIFIED = 0;
  NEW = 1;
  RECONCILED = 2;
  FAILED = 3;
  NO_CHANGES = 4;
}
```

diode-server/reconciler/logs_retriever.go
```diff
@@ -81,10 +82,59 @@ func decodeBase64ToInt64(encoded string) (int64, error) {
 	return num, nil
 }
 
+func retrieveIngestionStatsSummary(ctx context.Context, client RedisClient) (*reconcilerpb.RetrieveIngestionLogsResponse, error) {
+
+	pipe := client.Pipeline()
+
+	results := make([]*redis.Cmd, 0)
+	results = append(results, pipe.Do(ctx, "FT.SEARCH", "ingest-entity", "*", "LIMIT", 0, 0))
+	for i := 1; i < 5; i++ {
```
mfiedorowicz: Some comment would be useful here, what are we doing here and why 5? Assuming ingestion state `NEW`, we can also use existing code generate from proto `reconcilerpb.State_NO_CHANGES` (4)

diode-server/reconciler/logs_retriever.go
leoparente: cool. it fixed some files!

diode-proto/diode/v1/reconciler.proto
leoparente: hmm, the way I've implemented it was that if it is passed we only return the summary and return, no log returned.. 'include_metrics' seems to give the impression that we should return specified logs and also 'summary'

diode-server/reconciler/logs_retriever.go
leoparente: agree

diode-proto/diode/v1/reconciler.proto
leoparente: don't get the enum comparison

diode-proto/diode/v1/reconciler.proto
leoparente: yes, I'm fine changing it to either

diode-server/reconciler/logs_retriever.go
leoparente: agree to change it

diode-proto/diode/v1/reconciler.proto
mfiedorowicz: Got it, thanks for clarification, missing `only` in the comment tripped me over.  `metrics_only` sounds better?

diode-proto/diode/v1/reconciler.proto
mfiedorowicz: Ah, mea culpa: conflated enum iota values with field numbers here 🙈

diode-proto/diode/v1/reconciler.proto
leoparente: Changed to Ingestion Metrics

diode-proto/diode/v1/reconciler.proto
leoparente: done, removed optional

diode-proto/diode/v1/reconciler.proto
leoparente: changed it!

mfiedorowicz (approved):
🙇

PR #170, opened 2024-09-16
feat: add golangci-lint gci linter (go package import order)

mfiedorowicz:
Configuring golangci-lint gci linter: https://golangci-lint.run/usage/linters/#gci


leoparente (approved):
LGTM

PR #172, opened 2024-09-19
feat: store generated changeset into ingestion log

leoparente:
image


diode-server/reconciler/ingestion_processor.go
```diff
@@ -356,6 +376,24 @@ func normalizeIngestionLog(l []byte) []byte {
...
+	var brotliBuf bytes.Buffer
+	brotliWriter := brotli.NewWriter(&brotliBuf)
+	if _, err = brotliWriter.Write(csJSON); err != nil {
+		return "", fmt.Errorf("failed to compress changeset: %v", err)
+	}
+	if err = brotliWriter.Close(); err != nil {
+		return "", fmt.Errorf("failed to compress changeset: %v", err)
+	}
+
+	return base64.StdEncoding.EncodeToString(brotliBuf.Bytes()), nil
```
mfiedorowicz: Did we check the size of raw brotli bytes output vs brotli bytes encoded with base64?

diode-proto/diode/v1/reconciler.proto
```diff
@@ -70,6 +70,12 @@ message IngestionMetrics {
   int32 no_changes = 5;
 }
 
+// A change set
+message ChangeSet {
+  string id = 1; // A change set ID
+  string data = 2; // Base64 string representing the change set
```
mfiedorowicz: Could we ament comment for `data`, please? I'd suggest to use something generic, as data can be of any string form (raw JSON string or brotli or something else), e.g.
```suggestion
  string data = 2; // A change set data
```

diode-server/reconciler/ingestion_processor_internal_test.go
mfiedorowicz, line 1: Can we add tests for compressed change sets, please? E.g. to assert compression works as expected with `compressChangeSet` function.

leoparente: added it

mfiedorowicz (approved):
LGTM

PR #173, opened 2024-09-20
fix: extract ingested entity from proto

mfiedorowicz:
- use protobuf data instead of decoding with mapstructure (used mainly for json [stdlib] marshalled structs previously)
- add methods for converting protobuf entities into netbox corresponding structs
- adjust unit tests
- bonus: log retrieving by state as string instead of integer

Capture-2024-09-20-113607


leoparente (approved):
LGTM!

PR #175, opened 2024-09-24
feat: redis FT.SEARCH index adjustments

mfiedorowicz:
- add fields for producer app name and version
- for most of index fields use TAG as more suitable for filtering instead of TEXT (context: https://redis.io/docs/latest/develop/interact/search-and-query/advanced-concepts/escaping/)
- update querying and add escaping function


leoparente (approved):
Great. LGTM

PR #182, opened 2024-10-07
feat(proto): add entity_label field option

mfiedorowicz:
Add `entity_label` field option to indicate which entity field should be used as label.


leoparente (approved):
Mastering Protobuf. LGTM!

mfiedorowicz:
Seems to be redundant, closing

PR #193, opened 2024-11-20
fix: ingestion processor - continue processing ingestion logs on failure

mfiedorowicz:
- continue processing received ingestion logs on failure
- log errors


leoparente (approved):
LGTM

PR #194, opened 2024-11-25
feat: add migrator package

mfiedorowicz:
- add `migrator` package utilising https://github.com/pressly/goose for applying and rolling back database migrations against given database and with provided path to database migration files
- upgrade diode services go 1.23 and dependencies
- replace deprecated `grpc.DialContext` with `grpc.NewClient`


leoparente (approved):
LGTM!

PR #198, opened 2024-12-03
feat: introduce postgres dbstore with migrations

mfiedorowicz:
- add postgres db to docker compose
- add postgres schemas/migrations and initial queries
- generate dbstore code using sqlc
- add repositories for ingestion logs and change sets
- run migration on reconciler service startup

Screenshot 2024-12-03 at 14 30 52


diode-server/sqlc.yaml
```diff
@@ -0,0 +1,12 @@
+version: "2"
```
leoparente, line 1: sqlc runs transparently at build time? if not, should we add this to Makefile?

diode-server/reconciler/mocks/client.go
```diff
@@ -1,4 +1,4 @@
-// Code generated by mockery v2.43.2. DO NOT EDIT.
+// Code generated by mockery v2.42.1. DO NOT EDIT.
```
leoparente: maybe try to upgrade mockery to latest? but only cosmetic change

mfiedorowicz: That's odd, thought I was on the latest hence the change 🤨 Good spot

diode-server/sqlc.yaml
mfiedorowicz: sqlc being run locally similarly to mockery, will add recipes for both then

diode-server/netboxdiodeplugin/mocks/netboxapi.go
```diff
@@ -5,9 +5,8 @@ package mocks
 import (
 	context "context"
 
-	mock "github.com/stretchr/testify/mock"
```
leoparente, line 8: is this change correct?

leoparente (approved):
Overall, LGMT!

diode-server/netboxdiodeplugin/mocks/netboxapi.go
mfiedorowicz: That's what mockery generates

diode-server/sqlc.yaml
mfiedorowicz: Added following recipes:
```
make install-dev-tools
make gen-mocks
make gen-dbstore
```

leoparente: great! thanks for that

PR #202, opened 2024-12-13
feat: add branch support

ltucker:
Adds ability to specify netbox branch in applier/differ and netbox plugin client.
All automatic operations still default to main branch.

### DIFF
--- a/diode-server/netboxdiodeplugin/client.go
+++ b/diode-server/netboxdiodeplugin/client.go
@@ -289,7 +289,7 @@ func (c *Client) RetrieveObjectState(ctx context.Context, params RetrieveObjectS
 	}
 	branchID := strings.TrimSpace(params.BranchID)
 	if branchID != "" {
-		queryParams.Set(NetboxBranchParam, branchID)
+		queryParams.Set(NetBoxBranchParam, branchID)
 	}
 	for k, v := range params.Params {
 		queryParams.Set(k, v)
@@ -444,7 +444,7 @@ func (c *Client) ApplyChangeSet(ctx context.Context, payload ChangeSetRequest) (
 
 	branchID := strings.TrimSpace(payload.BranchID)
 	if branchID != "" {
-		req.Header.Set(NetboxBranchHeader, branchID)
+		req.Header.Set(NetBoxBranchHeader, branchID)
 	}
 
 	resp, err := c.httpClient.Do(req)
--- a/diode-server/netboxdiodeplugin/client.go
+++ b/diode-server/netboxdiodeplugin/client.go
@@ -15,6 +15,7 @@ import (
 	"os"
 	"reflect"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/mitchellh/mapstructure"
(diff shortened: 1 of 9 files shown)


diode-server/netboxdiodeplugin/client.go
```diff
@@ -42,6 +43,11 @@ const (
 	defaultBaseURL = "http://127.0.0.1:8080/api/plugins/diode"
 
 	defaultHTTPTimeoutSeconds = 5
+
+	// NetboxBranchHeader is an HTTP header that indicates the netbox branch to target
+	NetboxBranchHeader = "X-Netbox-Branch"
+	// NetboxBranchParam is a query parameter that indicates the netbox branch to target
+	NetboxBranchParam = "_branch"
```
mfiedorowicz: Let's try to keep consitency with `NetBox` name casing when possible, also it might matter in the header as well as should match https://github.com/netboxlabs/netbox-branching/blob/develop/netbox_branching/constants.py#L8

```suggestion
	// NetBoxBranchHeader is an HTTP header that indicates the NetBox branch to target
	NetBoxBranchHeader = "X-NetBox-Branch"
	// NetBoxBranchParam is a query parameter that indicates the NetBox branch to target
	NetBoxBranchParam = "_branch"
```

diode-server/netboxdiodeplugin/client_test.go
```diff
@@ -722,6 +773,11 @@ func TestApplyChangeSet(t *testing.T) {
 				assert.Equal(t, r.Header.Get("Authorization"), fmt.Sprintf("Token %s", tt.apiKey))
 				assert.Equal(t, r.Header.Get("User-Agent"), fmt.Sprintf("%s/%s", netboxdiodeplugin.SDKName, netboxdiodeplugin.SDKVersion))
 				assert.Equal(t, r.Header.Get("Content-Type"), "application/json")
+				if tt.changeSetRequest.BranchID != "" {
+					assert.Equal(t, r.Header.Get("X-Netbox-Branch"), tt.changeSetRequest.BranchID)
+				} else {
+					assert.Len(t, r.Header.Values("X-Netbox-Branch"), 0)
+				}
```
mfiedorowicz: Can we use `netboxdiodeplugin.NetBoxBranchHeader` constant instead of hardcoded values or do we want to assert if the client implementation (with constant) against to expected header name here?

diode-server/reconciler/differ/differ.go
```diff
@@ -30,7 +30,7 @@ type ObjectState struct {
 }
 
 // Diff compares ingested entity with the intended state in NetBox and returns a change set
-func Diff(ctx context.Context, entity IngestEntity, netboxAPI netboxdiodeplugin.NetBoxAPI) (*changeset.ChangeSet, error) {
+func Diff(ctx context.Context, entity IngestEntity, branch string, netboxAPI netboxdiodeplugin.NetBoxAPI) (*changeset.ChangeSet, error) {
```
mfiedorowicz: For consistency with applier, could we name the param `branchID` instead?

mfiedorowicz (commented):
Thanks Luke, just a few minor comments to consider, other than that functionally looks great!

diode-server/reconciler/differ/differ.go
ltucker: oh yeah good catch (fixed)

diode-server/netboxdiodeplugin/client_test.go
ltucker: (changed) that's fine too -- the only difference I think is that it assumes that the constants are spelled correctly (doesn't catch spelling NetBoxBranchParam `_barnch` by accident) -- but that is unlikely and easy to notice.

mfiedorowicz (approved):
LGTM, thanks Luke!

PR #203, opened 2024-12-16
feat: ActionIngestionLog rpc

ltucker:
proto for ActionIngestionLog rpc + generated stubs


diode-proto/diode/v1/reconciler.proto
```diff
@@ -111,10 +111,38 @@ message RetrieveIngestionLogsResponse {
...
+  string ingestion_log_id = 4;
+  optional string branch_id = 5;
+}
+
+// The response from the ActionIngestionLog request
+message ActionIngestionLogResponse {
+  message Error {
+    string message = 1;
+    int32 code = 2;
+  }
```
ltucker, line 133: The IngestionLog returned has its own set of errors, not sure what the shape of these errors is currently so I just tacked this in ...

diode-proto/diode/v1/reconciler.proto
```diff
@@ -111,10 +111,38 @@ message RetrieveIngestionLogsResponse {
   string next_page_token = 3; // Token for the next page of results, if any
 }
 
+// The request to take action on an ingestion log
+message ActionIngestionLogRequest {
+  message DiffAction {}
+  message ApplyAction {}
+  message RejectAction {}
```
ltucker: @mfiedorowicz you mentioned using a `oneof` structure to represent the available actions -- I imagine in part so that we can reserve space for varying arguments -- arguably the `branch_id` could belong only to DiffAction and ApplyAction only if we prefer?

ltucker:
I generated the server end of these on the PR so we can have a feel for what the shapes it creates in golang are.

diode-proto/diode/v1/reconciler.proto
mfiedorowicz: That's correct, so we can also keep ingestion log ID and branch ID in the root of the message and replace these empty action type messages with i.e. `enum ActionType` that would define supported action types, especially if the branch ID is optional and making a new message just to hold branch ID might be excessive.

diode-proto/diode/v1/reconciler.proto
mfiedorowicz: > A little copying is better than a little dependency

diode-proto/diode/v1/reconciler.proto
ltucker: That feels like the right idea for covering these to me -- if the operations were much more detailed I think we'd just want additional rpcs.

diode-server/gen/diode/v1/reconcilerpb/reconciler.pb.go
```diff
@@ -77,6 +77,58 @@ func (State) EnumDescriptor() ([]byte, []int) {
...
 }
 
+type ActionIngestionLogRequest_Action int32
+
+const (
+	ActionIngestionLogRequest_ACTION_UNSPECIFIED ActionIngestionLogRequest_Action = 0
+	ActionIngestionLogRequest_ACTION_DIFF        ActionIngestionLogRequest_Action = 1
+	ActionIngestionLogRequest_ACTION_APPLY       ActionIngestionLogRequest_Action = 2
+	ActionIngestionLogRequest_ACTION_REJECT      ActionIngestionLogRequest_Action = 3
+)
```
ltucker, line 87: feels a bit java-verbose here I think in the generation due to nesting+prefixing perhaps

mfiedorowicz: Indeed, we can also extract this enum definition outside of `ActionIngestionLogRequest` message and name it `ActionType` or something like that, then it would generate just `ActionType_*` constants

ltucker: yeah this is about as terse as we're likely to get from it I think.  I used `UNSPECIFIED_ACTION` here because it complained about UNSPECIFIED being a part of the State enum.

mfiedorowicz (approved):

PR #204, opened 2024-12-16
chore: enable gofumpt lint

ltucker:


ltucker:
The linter output in github is not very specific... basically just tells you if you're in violation or not ```Error: File is not `gofumpt`-ed (gofumpt)``` (though usually near the violation at least).

I added a target to fix lint suggestions in the local Makefile as well and allowed it to clean existing violations.

mfiedorowicz (approved):
👍

PR #205, opened 2024-12-17
feat: postgres datastore implementation - part 2

mfiedorowicz:
- consolidate ingestion logs and change sets repositories into single one (as related to each other)
- remove redis migrations
- remove writing ingestion logs to redis
- refactor postgres schemas
- implement postgres repository

Screenshot 2024-12-20 at 16 43 38

Integration tests:
Screenshot 2024-12-20 at 16 32 29


diode-server/reconciler/ingestion_processor.go
```diff
@@ -416,13 +386,13 @@ func (p *IngestionProcessor) CreateIngestionLogs(ctx context.Context, ingestReq
 			continue
 		}
 
-		ingestionLogID := ksuid.New().String()
+		ingestionLogKSUID := ksuid.New().String()
```
leoparente: [Question] Why we are using ksuid here and then we change in line 407 to uuid?

leoparente:
[Question] did you run go mod tidy after changes?

diode-server/reconciler/ingestion_processor.go
mfiedorowicz: Good spot, forgot to make clean up on redis related things.

diode-server/dbstore/postgres/repository.go
```diff
@@ -0,0 +1,255 @@
+package postgres
```
ltucker, line 1: the insides of this don't have much in the way of coverage, we could consider adding something like [embedded postgres](https://github.com/fergusstrange/embedded-postgres) to get more interesting tests of the system I think. It's not perfect / has quirks, but it is realistic :D 

may not be worthwhile immediately especially if you're considering changing the interface.

mfiedorowicz: Agree, coverage dropped overall by ~5%, to be addressed separately 👍

ltucker:
I don't see any showstoppers here, I've been using / looking over this branch for the last few days without issue.
Especially in the interest of not making this any larger, I'm fine with this on the feature branch and addressing additional things with followups.

ltucker (approved):

diode-server/reconciler/repository.go
```diff
@@ -0,0 +1,17 @@
...
+)
+
+// Repository is an interface for interacting with ingestion logs and change sets.
+type Repository interface {
+	CreateIngestionLog(ctx context.Context, ingestionLog *reconcilerpb.IngestionLog, sourceMetadata []byte) (*int32, error)
+	UpdateIngestionLogStateWithError(ctx context.Context, id int32, state reconcilerpb.State, ingestionError *reconcilerpb.IngestionError) error
+	RetrieveIngestionLogs(ctx context.Context, filter *reconcilerpb.RetrieveIngestionLogsRequest, limit int32, offset int32) ([]*reconcilerpb.IngestionLog, error)
+	CountIngestionLogsPerState(ctx context.Context) (map[reconcilerpb.State]int32, error)
+	CreateChangeSet(ctx context.Context, changeSet changeset.ChangeSet, ingestionLogID int32) (*int32, error)
+}
```
ltucker, line 17: discussed a bit offline, would like to see this continue to evolve, possibly bury dbids inside here and/or think about the structures that go in and out separately from the grpc interface and how to address transactionality.

PR #218, opened 2025-01-03
feat: add deviation name to change set

mfiedorowicz:
- deprecate data_type and substitute with object_type in ingestion log (for consistency)
- add deviation name to change set based on primary entity/object change
- psql: delete changes and change set on cascade


ltucker:
talked through in person

ltucker (approved):

PR #223, opened 2025-01-07
chore: ingestion refactoring

ltucker:
* encapsulate ingestion processor ops in an interface
* allow adjustment of migrations tracking table in migrator


mfiedorowicz (approved):
Looks great, thanks!

PR #241, opened 2025-01-27
fix: add custom json unmarshal for ip address assigned object

ltucker:
This fixes an error in `MarshalChangeDataToJSON` when working with an `ipam.ipaddress` containing an assigned object.

```failed to marshal after state: failed to unmarshal ip address data: json: cannot unmarshal object into Go struct field IpamIPAddress.assigned_object of type netbox.IPAddressAssignedObject ```

The AssignedObject field of IpamIPAddress (IPAddressAssignedObject) is an interface type which json cannot directly unmarshal (doesn't know which implementing type to create).  Elsewhere there is a mapstructure helper `IpamIPAddressAssignedObjectHookFunc` that is used. 

This adds a custom `UnmarshalJSON` to the `IpamIPAddress` struct which does something very similar.  Possibly could be combined or simplified eslewhere, though mapstructure is used directly.

### DIFF
--- a/diode-server/netbox/ipam_test.go
+++ b/diode-server/netbox/ipam_test.go
@@ -4,8 +4,9 @@ import (
 	"encoding/json"
 	"testing"
 
-	"github.com/netboxlabs/diode/diode-server/netbox"
 	"github.com/stretchr/testify/assert"
+
+	"github.com/netboxlabs/diode/diode-server/netbox"
 )
 
 func TestIpamIPAddress_UnmarshalJSON(t *testing.T) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by leoparente — APPROVED** [review]
great fix. LGTM

File: diode-server/netbox/ipam_test.go [hunk 1]
```diff
@@ -0,0 +1,86 @@
+package netbox
```
leoparente: question. cant it be ipam_test / netbox_test (external package)?

ltucker: good catch

PR #249, opened 2025-02-18
fix: trigger update to updated_at on ingestion log row update

ltucker:


mfiedorowicz (approved):

mfiedorowicz:
Could we also add the trigger to `deviation_types.updated_at` (here or in a separate PR)?

ltucker:
yes, the function can be used elsewhere -- will have to be separate PR though (deviation_types doesn't exist here)

PR #255, opened 2025-04-02
adds rate limiting to calls to netbox

jajeffries:
## In scope of this PR
Adds rate limiting for request from diode-server to netbox. Set based on the environment variables
- `DIODE_TO_NETBOX_RATE_LIMITER_RPS` - defaulted to 20
- `DIODE_TO_NETBOX_RATE_LIMITER_BURST` - defaulted to 1

## Out of scope of this PR
- retrying due to timeouts

## Future changes
- to support multitenancy a separate client could be created per netbox instance with different environment

## Questions
- Given this PR introduces a breaking change to the netboxdiodeplugin.NewClient function, what does that mean for versioning?

### DIFF
--- a/diode-server/netboxdiodeplugin/client.go
+++ b/diode-server/netboxdiodeplugin/client.go
@@ -17,6 +17,7 @@ import (
 	"strings"
 	"time"
 
+	"golang.org/x/time/rate"
 	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/proto"
 
@@ -160,6 +161,7 @@ type Client struct {
 	logger     *slog.Logger
 	httpClient *http.Client
 	baseURL    *url.URL
+	limiter    *rate.Limiter
 }
 
 // NewHTTPTransport creates a http Transport Layer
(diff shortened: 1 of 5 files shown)


diode-server/netboxdiodeplugin/client.go
```diff
@@ -182,7 +184,7 @@ func NewHTTPTransport() *http.Transport {
 }
 
 // NewClient creates a new NetBox Diode plugin client
-func NewClient(logger *slog.Logger, apiKey string) (*Client, error) {
+func NewClient(logger *slog.Logger, apiKey string, rateLimitRps, rateLimitBurstRps int) (*Client, error) {
```
ltucker, line 187: There is some usage of this client elsewhere, but I don't think there is a versioning concern at this point.  When we adjust those peices I think we'd also want to consider whether there are separate limits (or no limits) on usage triggered interactively (vs this automated pipeline)

ltucker (approved):

diode-server/netboxdiodeplugin/client.go
mfiedorowicz: We could consider here variadic functions with optional params (which is a nice pattern)

jajeffries: Yes, I considered variadic params, but worried that the multiple levels of defaults (in the config and here) could get confusing. I suppose you could say if the rate limit params aren't passed then dont rate limit, but that might be confusing if you've set the env vars and expect requests to be rate limited.

RangerRick: Not to butt in, but when you start getting too many things (especially optional) and you don't know if it will have to expand even further in the future, you might just want to make a simple config object that you pass in...

Once you're starting to worry about potential future versioning, that smells like variadic functions are just asking for trouble. :D

mfiedorowicz (approved):

mfiedorowicz (approved):

PR #256, opened 2025-04-03
fix: include several missing many-many references in data model

ltucker:
updates protobuf to include several missing many-many references eg Site.asns


diode-server/gen/netbox/server_utils.go
```diff
@@ -554,46 +554,42 @@ func GetObjectTypeName(objectType string) (string, error) {
 
 func GetPrimaryValue(entity *pb.Entity) (string, error) {
     switch e := entity.GetEntity().(type) {
-    case *pb.Entity_AsnRange:
-        return fmt.Sprintf("%v", e.AsnRange.Name), nil
     case *pb.Entity_Asn:
         return fmt.Sprintf("%v", e.Asn.Asn), nil
-    case *pb.Entity_CircuitGroup:
-        return fmt.Sprintf("%v", e.CircuitGroup.Name), nil
+    case *pb.Entity_AsnRange:
```
ltucker, line 559: I changed this to a stable ordering (I think), this is mostly just shuffling into alphabetical.

mfiedorowicz (approved):

diode-proto/diode/v1/ingester.proto
```diff
@@ -14,7 +14,10 @@ option go_package = "github.com/netboxlabs/diode-sdk-go/diode/v1/diodepb";
 // An ingest entity wrapper
 
 message Entity {
-  google.protobuf.Timestamp timestamp = 1 [(validate.rules).timestamp.required = true, (validate.rules).timestamp.lt_now = true];
+  google.protobuf.Timestamp timestamp = 1 [
```
ltucker, line 17: interestingly, it looks like buf reformatted this when generating? possibly preferable?

ltucker: maybe need a way to do this update so it always does this ...

mfiedorowicz: for readability 👍

PR #263, opened 2025-04-09
feat: updates for custom fields

ltucker:
updates protobufs to accept updated custom field value types


diode-server/gen/diode/v1/diodepb/ingester.pb.validate.go
```diff
@@ -17522,6 +17522,59 @@ func (m *CustomFieldValue) validate(all bool) error {
 			}
 		}
 
+	case *CustomFieldValue_Json:
```
jajeffries: It looks like there are a few [custom field types ](https://netboxlabs.com/docs/netbox/en/stable/customization/custom-fields/) that aren't in this switch statement. Specifically
- Long text
- Selection
- Multiple selection
- Multiple object

Is there a reason for this or are we just not supporting them for now?

mfiedorowicz (approved):

diode-server/gen/diode/v1/diodepb/ingester.pb.validate.go
ltucker: I haven't differentiated text from long text, url, choice etc on the protobuf side.  It will accept text into long text type field or text into a url field etc.  I think it could be nice for developer ergonomics to just line them up and have them all enumerated though so it's clear what's valid / potentially provide validation.


Mutliple value things are not a part of the oneof "value" in protobuf (can't have a repeated value in a oneof in protobuf I believe).  There are sidecar fields in the CustomFieldValue message called ... string_array and object_array I believe to cover mutli value inputs.  These could also be renamed to more closely match the target types like multiple_selections / multiple_objects in the protobuf.

ltucker: updated to include a field for each / adjusted naming to be closer to netbox naming.

PR #264, opened 2025-04-10
feat: check for breaking protopub changes

jajeffries:
Adds:
- github action to check for breaking changes - this currently just runs manually but we can change this in future we have stabilised the protobuf definition
- a make file recipe to check for breaking changes locally as part of development

Both of these compare against develop to work out what is a breaking change.

### DIFF
--- a/.github/workflows/protobuf-break-detection.yaml
+++ b/.github/workflows/protobuf-break-detection.yaml
@@ -1,0 +1,25 @@
+name: Buf Breaking Change Check
+
+on:
+  workflow_dispatch:
+
+jobs:
+  buf-breaking-check:
+    name: Buf Breaking Check
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/protobuf-break-detection.yaml
```diff
@@ -0,0 +1,25 @@
+name: Buf Breaking Change Check
+
+on:
+  workflow_dispatch:
```
mfiedorowicz, line 4: Can we trigger this only on diode-proto dir contents changes?

jajeffries: @mfiedorowicz Yes, do we want that to happen while we're still working on this? If so I think it may break on any changes until we merge into develop.

mfiedorowicz: As it's `workflow_dispatch` only we can leave as is after all for time being 👍

mfiedorowicz (approved):

PR #266, opened 2025-04-11
fix: fix recording of errors from diffs that generate a 500 from NetBox

ltucker:


diode-server/netboxdiodeplugin/client_test.go
```diff
@@ -484,6 +549,7 @@ func TestApplyChangeSet(t *testing.T) {
 					},
 				},
 			},
+			mockStatusCode:   http.StatusBadRequest,
```
ltucker, line 552: drive-by-fix this was quietly panic-ing but only printed when a test is failing.   providing a status here doesn't alter the test afaict.

```
2025/04/10 20:31:46 http: panic serving 127.0.0.1:51277: invalid WriteHeader code 0
goroutine 122 [running]:
net/http.(*conn).serve.func1()
	/usr/local/go/src/net/http/server.go:1947 +0xe4
panic({0x103360b60?, 0xc00018aa20?})
	/usr/local/go/src/runtime/panic.go:787 +0x124
net/http.checkWriteHeaderCode(...)
```

diode-server/netboxdiodeplugin/client.go
```diff
@@ -302,16 +302,16 @@ func (c *Client) GenerateDiff(ctx context.Context, payload GenerateDiffRequest)
 
 	c.logger.Debug("generate diff", "statusCode", resp.StatusCode, "response", string(respBytes))
 
-	// return errors with 4xx status code
-	if resp.StatusCode >= http.StatusBadRequest {
-		return nil, changeset.NewError("generate diff failed", diodeErrors.ErrCodeOpsGenerateDiff, respBytes)
-	}
-
```
ltucker, line 309: if it returns respBytes inside a changeset.Error here, the body hasn't yet been successfully parsed as json. So a 500 will generate a changset.Error that has invalid json as the details and fails to write to the database afaict.  Changed to return after (like apply), which will send back json parsing failures as a "failed to unmarshal" error / handled as internal error instead.

diode-server/reconciler/ops.go
```diff
@@ -127,6 +129,6 @@ func handleChangeSetError(err error) error {
 	return &changeset.Error{
 		Message: err.Error(),
 		Code:    diodeErrors.ErrCodeInternal,
-		Details: []byte{},
+		Details: []byte(genericInternalServerErrorDetails),
 	}
```
ltucker: empty bytes fails json.Marshal in the repository.  we could allow that in the repository additionally, but I just put a generic error message in this particular case.  can be whatever.

ltucker: If we prefer no details for this `[]byte("null")` also works here ...

mfiedorowicz: Adding `omitempty` into https://github.com/netboxlabs/diode/blob/feat-postgres-implementation-and-enhancements/diode-server/reconciler/changeset/changeset.go#L69 would solve empty bytes issue I believe

ltucker: sure, added.

mfiedorowicz (approved):

PR #267, opened 2025-04-11
feat(diode-auth): implement /token endpoint

jajeffries:
Adds introspect endpoint and adds fixes for running on linux


diode-server/auth/server.go
```diff
@@ -2,25 +2,41 @@ package auth
...
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"log/slog"
 	"net/http"
 	"time"
 
+	"github.com/MicahParks/keyfunc"
+	"github.com/golang-jwt/jwt/v4"
```
mfiedorowicz: Anything behind using v4 instead of newer v5?

diode-server/docker/docker-compose.yaml
```diff
@@ -152,7 +155,8 @@ services:
       - HYDRA_ADMIN_URL=http://diode-hydra:4445
     volumes:
       - ./oauth2/bootstrap-clients.sh:/etc/config/oauth2/bootstrap-clients.sh:z,ro
-      - ./oauth2/client:/etc/config/oauth2/client:z
+      - ./oauth2/client:/etc/config/oauth2/client
+    user: "appuser:appgroup"
```
mfiedorowicz: Is it still needed if we use this user in docker image by default?

diode-server/auth/tokens.go
```diff
@@ -0,0 +1,16 @@
+package auth
+
+import "github.com/golang-jwt/jwt/v4"
```
mfiedorowicz: Ditto

diode-server/auth/server.go
jajeffries: I can't get the keyfunc module to work with v5

diode-server/auth/tokens.go
jajeffries: See https://github.com/netboxlabs/diode/pull/267#discussion_r2041791843

diode-server/auth/server.go
mfiedorowicz: That's odd cause keyfunc itself depends on v5 https://github.com/MicahParks/keyfunc/blob/main/go.mod#L7

jajeffries: Upgraded to v5. I had the wrong version of keyfunc

leoparente (approved):

mfiedorowicz (approved):
🚀

PR #268, opened 2025-04-12
fix: use snake_case protoJSON

ltucker:
outputs protoJSON using snake_case (preserve protobuf field names)


ltucker:
no rush on this end since other end will accept both / display as snake_case, but seems easier to think about if it's actually same all the way through.

leoparente (approved):
LGTM

mfiedorowicz (approved):
👍

PR #299, opened 2025-04-28
chore: remove netbox docker-compose

mfiedorowicz:
Not needed anymore here as we have dedicated netbox docker compose in diode-netbox-plugin repository


leoparente (approved):
LGTM

PR #327, opened 2025-05-13
fix: helm chart values small fixes

RangerRick:
* some boolean values passed to the diode reconciler end up in the environment and thus should be treated as strings in `values.yaml`
* `postgresql:` is used in values and templates, but the chart dependency only checks for `postgres.enabled`


mfiedorowicz (approved):
LGTM, thanks for contribution!

PR #348, opened 2025-05-27
fix: extraHttpPaths is a list, not a dictionary

RangerRick:
This resolves the following Helm warning when setting paths:

coalesce.go:298: warning: cannot overwrite table with non table for diode.ingressNginx.extraHttpPaths (map[])
coalesce.go:298: warning: cannot overwrite table with non table for diode.ingressNginx.extraHttpPaths (map[])
coalesce.go:298: warning: cannot overwrite table with non table for diode.ingressNginx.extraHttpPaths (map[])

### DIFF
--- a/charts/diode/.gitignore
+++ b/charts/diode/.gitignore
@@ -1,0 +1,1 @@
+/charts/*.tgz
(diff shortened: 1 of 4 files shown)


mfiedorowicz (commented):
Can we bump the helm chart version and regenerate readme, please?

mfiedorowicz (approved):
Thank you

PR #350, opened 2025-05-29
chore: add helm dependencies without changing Github actions

ckdarby:
Saw that @mfiedorowicz had to fix the Github actions via 7aa77e3. Improvement, no need to muck around with the Github action when dealing with helm dependency.

**Changes**
- Locked action to the latest released tag 
- Using yq to parse out the yaml
- bash to send in name + repo URL to the helm repo add CLI.

Did not bump chart version. No changes in the chart itself, but that means the action won't run when merged.

The workflow might want to be modified in the future to enable manual runs of the GitHub Action to validate everything works without triggering a publish or release.

I manually bumped Chart version in my fork to confirm the Github action.

 Thoughts?


mfiedorowicz (approved):
LGTM, thanks for useful improvement 👍

PR #351, opened 2025-05-30
chore: add helm chart testing action

ckdarby:
@mfiedorowicz Further expanding some of the Github workflow for helm.

Want to introduce helm chart tester as a step towards eventually adding helm tests and bringing Kind as an action to test a real deployment of the chart.

Version bump in the Chart to meet the requirement of testing looking for a version increment bump if chart.yaml is modified and I needed to modified to keep ct linter happy with it's default.

### DIFF
--- a/.github/workflows/configs/helm-ct.yaml
+++ b/.github/workflows/configs/helm-ct.yaml
@@ -1,0 +1,2 @@
+validate-maintainers: false
+skip-helm-dependencies: true
--- a/.github/workflows/helm-lint.yaml
+++ b/.github/workflows/helm-lint.yaml
@@ -4,7 +4,11 @@ on:
   push:
     branches:
       - "!release"
+    paths:
+      - "charts/**"
   pull_request:
+    paths:
+      - "charts/**"
 
 concurrency:
   group: ${{ github.workflow }}
--- a/.github/workflows/helm-lint.yaml
+++ b/.github/workflows/helm-lint.yaml
@@ -1,13 +1,10 @@
 name: Helm - lint
+
 on:
   push:
     branches:
       - "!release"
-    paths:
-      - "charts/**"
   pull_request:
-    paths:
-      - "charts/**"
 
 concurrency:
   group: ${{ github.workflow }}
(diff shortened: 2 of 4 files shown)


ckdarby:
Any reviewer, please chime in if there are changes to follow conventions. Tried my best to follow what I thought would make the most sense introducing this.

Once this MR is merged I'd like to introduce [kind](https://github.com/helm/kind-action) to fully test the helm chart in a functional k8s cluster to end-to-end test.

.github/workflows/helm-lint.yaml
```diff
@@ -1,13 +1,10 @@
...
+
 on:
   push:
     branches:
       - "!release"
-    paths:
-      - "charts/**"
   pull_request:
-    paths:
-      - "charts/**"
```
ckdarby: Path is gone because ct list-changed step handles this now.

mfiedorowicz: These paths help limit when to trigger and run GHA workflow/workers so would like to keep these, otherwise we would be running these each time server changes but chart changes not so often.

ckdarby: True. Can adjust.

Outside of the scope of this conversation - Has anyone discussed moving out the Netbox related (sure it isn't just Diode) to a mono chart repo?

.github/workflows/configs/helm-ct.yaml
```diff
@@ -0,0 +1,2 @@
+validate-maintainers: false
```
ckdarby, line 1: FYI: This was added because the Chart maintainer name is "NetBox Labs" and chart test does a validation to see if a github org/user exists matching the name which it returns 404 with the current one.

.github/workflows/helm-lint.yaml
ckdarby: @mfiedorowicz Changed to include the `paths` again.

mfiedorowicz: Thanks @ckdarby. Will do full review early next week.

.github/workflows/helm-lint.yaml
```diff
@@ -20,16 +21,44 @@ permissions:
...
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
-      - name: Setup Helm
-        uses: azure/setup-helm@fe7b79cd5ee1e45176fcad797de68ecaf3ca4814 # v4
-      - name: Update helm dependencies
-        run: helm dependency update diode
-      - name: Run helm lint
-        run: helm lint diode
+        uses: actions/checkout@v3
```
mfiedorowicz: Can we keep using latest `v4` as is?

.github/workflows/helm-lint.yaml
```diff
@@ -20,16 +21,44 @@ permissions:
...
+        with:
+          fetch-depth: 0
+
+      - name: Set up Helm
+        uses: azure/setup-helm@v4.2.0
```
mfiedorowicz: We should pin to commit hash as this action is not immutable

charts/diode/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: diode
 description: A Helm chart for Diode
 type: application
-version: 1.4.2
+version: 1.4.3
```
mfiedorowicz: Outdated already due to other merge with version bump to 1.5.0

.github/workflows/helm-lint.yaml
ckdarby: Yes, I apologize, it was based on the example from the chart testing that I referenced.

ckdarby:
@mfiedorowicz All completed.

mfiedorowicz (approved):
Looks good, thanks @ckdarby!

mfiedorowicz:
> Any reviewer, please chime in if there are changes to follow conventions. Tried my best to follow what I thought would make the most sense introducing this.
> 
> Once this MR is merged I'd like to introduce [kind](https://github.com/helm/kind-action) to fully test the helm chart in a functional k8s cluster to end-to-end test.

Re kind for testing end-to-end, could you please issue a feature request first with detailed approach you'd like to suggest?

PR #352, opened 2025-05-30
feat: add support for supplying additional environment variables

RangerRick:
This makes it simpler to inject environment that might be coming from multiple configuration/secrets/etc. without trying to runtime-inject data into secrets.

### DIFF
--- a/charts/diode/Chart.lock
+++ b/charts/diode/Chart.lock
@@ -1,4 +1,7 @@
 dependencies:
+- name: common
+  repository: https://charts.bitnami.com/bitnami
+  version: 2.31.1
 - name: redis
   repository: https://charts.bitnami.com/bitnami
   version: 20.11.5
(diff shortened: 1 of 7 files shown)


mfiedorowicz (approved):

ckdarby:
Question, I saw that `extraHttpPaths` is an `[]` and was handled differently in the yaml via range, [here](https://github.com/netboxlabs/diode/blob/develop/charts/diode/templates/ingress-http.yaml#L46-L53)

Was there a reason for the difference with `extraEnvs`?

RangerRick:
> Question, I saw that `extraHttpPaths` is an `[]` and was handled differently in the yaml via range, [here](https://github.com/netboxlabs/diode/blob/develop/charts/diode/templates/ingress-http.yaml?rgh-link-date=2025-05-30T22%3A07%3A17Z#L46-L53)
> 
> Was there a reason for the difference with `extraEnvs`?

@ckdarby Good question!

The biggest advantage of using the bitnami common libraries is that they let you provide a string as an alternate value instead of whatever type it normally is, and it will render template controls at runtime. Not strictly _necessary_ but a useful flexible thing for volumes/envs/etc. so you can do more advanced stuff from parent charts.

So if the parent has extra stuff it wants to set conditionally, you can do:
```
extraEnvs: |-
  {{- if .Values.something.enabled }}
  - name: CUSTOM_ENV
    valueFrom:
      secretKeyRef:
        name: "my-secret"
        key: "key-name"
  {{- end }}
```

Probably worth eventually making another PR to normalize these to all use common rendering, but I was trying not to make this a giant refactor patch.

PR #356, opened 2025-06-06
chore: update readme and new quick start guide

rboucher-me:


GET_STARTED.md
mfiedorowicz, line 1: This duplicates instructions provided in https://github.com/netboxlabs/diode/blob/develop/diode-server/README.md. If we keep both, we introduce easy path to drift between these two so I'd recommend to keep one of them for consistency.

rboucher-me: Agreed on the duplication, but it also consolidates instructions from three different places (and those instructions are confusing in places). I'm looking for a short term fix to the problem of people not being able to get started with Diode. I suggest we merge this in and work towards deprecating it by cleaning up the rest of the docs. Thoughts?

mfiedorowicz (approved):

PR #362, opened 2025-06-11
feat: add hook so the auth bootstrap job re-runs on deployment

RangerRick:


mfiedorowicz (commented):
Chart version also needs update (detected by linter)

PR #367, opened 2025-06-20
fix: the filenames don't change, so don't parameterize them

RangerRick:
If you include diode using an `alias` in your `Chart.yaml`, the template checksums won't render because the filenames are, eg, `diode-auth-configmap.yaml` but it would try checksumming `<alias>-auth-configmap.yaml` instead.


mfiedorowicz (approved):

jajeffries (approved):
LGTM

PR #383, opened 2025-07-11
feat: basic deduplication of ingested entities

ltucker:
- deduplication of ingested entities


mfiedorowicz (approved):
LGTM, one ask to check if logs in CreateIngestionLog method are sufficient

diode-server/dbstore/postgres/migrations/20250707184637_entity_deduplication.sql
```diff
@@ -0,0 +1,18 @@
...
+ALTER TABLE ingestion_logs
+    ADD COLUMN entity_hash VARCHAR(64),
+    ADD COLUMN last_seen TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP,
+    ADD COLUMN duplicate_count INTEGER DEFAULT 0 NOT NULL;
+
+CREATE INDEX idx_ingestion_logs_entity_hash ON ingestion_logs(entity_hash);
+
+-- +goose Down
+
+
```
mfiedorowicz: nit: can we remove redundant new line?

diode-server/dbstore/postgres/queries/change_sets.sql
```diff
@@ -11,3 +11,15 @@ INSERT INTO changes (external_id, change_set_id, change_type, object_type, objec
                      sequence_number)
 VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12)
 RETURNING *;
+
+
```
mfiedorowicz: nit: redundant new line

diode-server/dbstore/postgres/queries/ingestion_logs.sql
```diff
@@ -42,3 +42,29 @@ WHERE (v_deviations.state = sqlc.narg('state') OR sqlc.narg('state') IS NULL)
...
 ORDER BY v_deviations.id DESC
 LIMIT sqlc.arg('limit') OFFSET sqlc.arg('offset');
+
+-- name: FindPriorIngestionLogByEntityHash :one
+WITH latest_change_sets AS (
+    SELECT DISTINCT ON (ingestion_log_id) 
+        ingestion_log_id, 
+        branch_id
+    FROM change_sets
+    ORDER BY ingestion_log_id, id DESC
```
mfiedorowicz, line 52: Do we need to index ingestion_log_id and id on change_sets table?

diode-server/dbstore/postgres/queries/ingestion_logs.sql
```diff
@@ -42,3 +42,29 @@ WHERE (v_deviations.state = sqlc.narg('state') OR sqlc.narg('state') IS NULL)
...
+)
```
mfiedorowicz, line 53: This CTE is going to scan whole change_sets table on each query whenever we handle new ingestion log, hence its performance can degrade with larger data sets (basic analysis on ~300k rows = ~200ms).
We can try to narrow down the scan by filtering early here, e.g.:
```sql
WITH latest_change_sets AS (
    SELECT DISTINCT ON (cs.ingestion_log_id)
        cs.ingestion_log_id,
        cs.branch_id
    FROM change_sets cs
    WHERE cs.ingestion_log_id IN (
        SELECT id FROM ingestion_logs
        WHERE entity_hash = sqlc.arg('entity_hash')
    )
    ORDER BY cs.ingestion_log_id, cs.id DESC
)
```

mfiedorowicz (commented):
Looks solid overall, one important query related comment worth considering.

diode-server/dbstore/postgres/queries/ingestion_logs.sql
ltucker: It's already indexed I believe.  from `00002_change_sets.sql`:

```CREATE INDEX IF NOT EXISTS idx_change_sets_ingestion_log_id ON change_sets (ingestion_log_id);```

diode-server/dbstore/postgres/queries/ingestion_logs.sql
ltucker: yeah seem unnecessary in total tbh -- should be structured like other queries that grab latest change set -- will take a look.

ltucker: I reworked this to just be a join which should be much better I think .. but you can give it a whirl on the actual dataset.

mfiedorowicz (approved):

diode-server/dbstore/postgres/queries/ingestion_logs.sql
mfiedorowicz: Apologies, CTE is actually performant, tested partially only contents of CTE 🙈

mfiedorowicz (approved):

PR #390, opened 2025-08-07
fix: Resilient redis client

MicahParks:
When Redis goes offline then comes back online, the reconciler does not automatically reconnect and read from the relevant stream as expected. These changes should have the reconciler automatically reconnect and read from the relevant stream.

### DIFF
--- a/diode-server/reconciler/ingestion_processor.go
+++ b/diode-server/reconciler/ingestion_processor.go
@@ -167,6 +167,7 @@ func (p *IngestionProcessor) consumeIngestionStream(ctx context.Context, redisSt
 				continue
 			}
 		}
+		p.logger.WarnContext(ctx, "Read from Redis stream.")
 		for _, msg := range streams[0].Messages {
 			_, err := p.handleStreamMessage(ctx, msg)
 			if err != nil {
--- a/diode-server/reconciler/ingestion_processor.go
+++ b/diode-server/reconciler/ingestion_processor.go
@@ -7,6 +7,7 @@ import (
 	"log/slog"
 	"os"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
@@ -158,6 +159,7 @@ func (p *IngestionProcessor) consumeIngestionStream(ctx context.Context, redisSt
 				}
 				cancel()
 			}
+			p.logger.WarnContext(ctx, "Redis stream client ping")
 		}
 	}()
 
@@ -179,6 +181,12 @@ func (p *IngestionProcessor) consumeIngestionStream(ctx context.Context, redisSt
 			p.logger.WarnContext(ctx, "Failed to read from Redis stream.",
 				"error", err,
 			)
+			if strings.Contains(err.Error(), "NOGROUP") {
+				err := p.redisStreamClient.XGroupCreateMkStream(ctx, redisStreamID, redisConsumerGroup, "$").Err()
+				if err != nil && err.Error() != RedisConsumerGroupExistsErrMsg {
+					p.logger.Debug("Failed to recreate Redis consumer group.")
+				}
+			}
 			select {
 			case <-ctx.Done():
 				p.logger.Debug("ingestion processor exiting consumer loop on request")
--- a/diode-server/reconciler/ingestion_processor.go
+++ b/diode-server/reconciler/ingestion_processor.go
@@ -8,7 +8,9 @@ import (
 	"os"
 	"strconv"
 	"sync"
+	"time"
 
+	"github.com/cloudflare/backoff"
 	"github.com/google/uuid"
 	"github.com/redis/go-redis/v9"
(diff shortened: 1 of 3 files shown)


diode-server/reconciler/ingestion_processor.go
```diff
@@ -153,8 +157,22 @@ func (p *IngestionProcessor) consumeIngestionStream(ctx context.Context, redisSt
...
+					p.logger.Debug("Failed to recreate Redis consumer group.")
+				}
+			}
+			select {
+			case <-ctx.Done():
+				p.logger.Debug("ingestion processor exiting consumer loop on request")
+				return nil
+			case <-time.After(b.Duration()):
+				continue
+			}
```
MicahParks, line 172: This backoff isn't strictly necessary and can be removed if desired.

diode-server/reconciler/ingestion_processor.go
```diff
@@ -153,8 +157,22 @@ func (p *IngestionProcessor) consumeIngestionStream(ctx context.Context, redisSt
 			Count:    100,
 		}).Result()
 		if err != nil || len(streams) == 0 {
-			continue
+			if strings.Contains(err.Error(), "NOGROUP") {
+				err := p.redisStreamClient.XGroupCreateMkStream(ctx, redisStreamID, redisConsumerGroup, "$").Err()
```
MicahParks, line 161: The main issue appears to be that the consumer group is not automatically recreated when a non-persistent Redis instance is used.

The writer to this stream should automatically create the stream itself when [`XADD`](https://redis.io/docs/latest/commands/xadd/) is called.

MicahParks: Is it practical to remove the consumer group requirement? Removing the consumer group requirement may make this fix unnecessary.

diode-server/reconciler/ingestion_processor.go
```diff
@@ -153,8 +157,22 @@ func (p *IngestionProcessor) consumeIngestionStream(ctx context.Context, redisSt
 			Count:    100,
 		}).Result()
 		if err != nil || len(streams) == 0 {
-			continue
+			if strings.Contains(err.Error(), "NOGROUP") {
```
MicahParks, line 160: This implementation checks to see if `NOGROUP` is in the error string before recreating the stream and group. This implementation means that we must encounter the error before ensuring the stream and consumer group exist.

Alternatively, this could be done in a separate goroutine.

MicahParks:
The way the `ingester` is currently written, it will to drop messages to the stream with an log statement and metric increment when Redis is down.

Should we write messages to a in-memory buffer and automatically retry them when the first write to Redis fails?

https://github.com/netboxlabs/diode/blob/1b8e9892faa2390685ddba618f1750104206d901/diode-server/ingester/component.go#L169-L179

diode-server/reconciler/ingestion_processor.go
```diff
@@ -153,8 +157,22 @@ func (p *IngestionProcessor) consumeIngestionStream(ctx context.Context, redisSt
...
+				err := p.redisStreamClient.XGroupCreateMkStream(ctx, redisStreamID, redisConsumerGroup, "$").Err()
+				if err != nil && err.Error() != RedisConsumerGroupExistsErrMsg {
+					p.logger.Debug("Failed to recreate Redis consumer group.")
+				}
+			}
```
MicahParks, line 165: By not adding a `continue` here, this is subject to the backoff duration. With the current backoff configuration, it could take a maximum of about 20 seconds between when Redis comes back online and when the consume group is created and when the stream is read from.

Adding a `continue` here should bring the maximum down to about 10 seconds, however that implementation would be subject to busy-waiting if recreating `XGroupCreateMkStream` consistently errors.

diode-server/reconciler/ingestion_processor.go
leoparente: I personally like a backoff approach

leoparente (approved):
Overall LGTM

diode-server/reconciler/ingestion_processor.go
mfiedorowicz: Consumer group is in place to ensure the message is consumer by one member of the consumer group only.

mfiedorowicz (approved):
Thanks!

mfiedorowicz:
> The way the `ingester` is currently written, it will to drop messages to the stream with an log statement and metric increment when Redis is down.
> 
> Should we write messages to a in-memory buffer and automatically retry them when the first write to Redis fails?
> 
> https://github.com/netboxlabs/diode/blob/1b8e9892faa2390685ddba618f1750104206d901/diode-server/ingester/component.go#L169-L179

This is good suggestion we should consider implementing as a separate pull request. Another thing it highlights though is that when adding message to redis stream fails, we log it as an error but don't return error in the response.

PR #397, opened 2025-08-21
fix: return gRPC error when failure to write to Redis stream

MicahParks:
This pull request returns a gRPC error to the requester when it fails to write the ingestion log to the Redis stream.

Clients should retry with backoff if this occurs.

The error returned should be the [`INTERNAL` status code](https://grpc.io/docs/guides/status-codes/#the-full-list-of-status-codes).

### DIFF
--- a/diode-server/ingester/component.go
+++ b/diode-server/ingester/component.go
@@ -174,7 +174,7 @@ func (c *Component) Ingest(ctx context.Context, in *diodepb.IngestRequest) (*dio
 	}).Err(); err != nil {
 		c.metrics.RecordIngestRequest(ctx, false)
 		c.logger.Error("failed to add element to the stream", "error", err, "streamID", streamID, "value", msg)
-		return nil, status.Error(codes.Internal, "unable to accept")
+		return nil, status.Error(codes.Internal, "")
 	} else {
 		entityCount := int64(len(in.GetEntities()))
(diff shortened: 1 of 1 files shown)


diode-server/ingester/component.go
```diff
@@ -166,18 +168,20 @@ func (c *Component) Ingest(ctx context.Context, in *diodepb.IngestRequest) (*dio
...
-	}).Err(); err != nil {
+	}).Err()
+	if err != nil {
 		c.metrics.RecordIngestRequest(ctx, false)
 		c.logger.Error("failed to add element to the stream", "error", err, "streamID", streamID, "value", msg)
-	} else {
-		entityCount := int64(len(in.GetEntities()))
-		c.metrics.RecordIngestRequest(ctx, true)
-		c.metrics.RecordIngestEntities(ctx, entityCount)
+		return nil, status.Error(codes.Internal, "")
```
MicahParks, line 178: Perhaps we should use a code reserved for gRPC users? Like `ABORTED`?

https://grpc.io/docs/guides/status-codes/#full-list-of-status-codes

I don't think it matters much. Perhaps `INTERNAL` is the best for this use case.

mfiedorowicz: `INTERNAL` seems appropriate for the use case 👍

mfiedorowicz (approved):

leoparente (approved):

PR #404, opened 2025-09-16
feat: improve helm chart postgresql secret handling

ecdobry:
The intention behind this pull request is to get diode's chart more inline with how the netbox community chart handles postgresql secrets. While not exactly the same, the goal is to create some familiarity between the two. 

Additionally if a consumer of this chart were to leverage an external postgresql database, they would need to override environment variables to do so. This is less than ideal and these changes provide an easier mechanism to use external databases.

`values.yaml`:
- Expanded `externalPostgresql` values: Added support for database, username, password, existingSecretName, and existingSecretKey fields in values.yaml

`_helpers.tpl`:
- `diode.postgresql.database` - Returns database name (built-in or external)
- `diode.postgresql.username` - Returns username (built-in or external)
- `diode.postgresql.secretname` - Returns appropriate secret name based on configuration
- `diode.postgresql.secretkey` - Returns appropriate secret key based on configuration

`diode-secrets.yaml`: creates external postgresql secrets when needed

`diode-reconciler-deployment.yaml`: properly references postgresql passwords from secrets

`diode-reconciler-configmap.yaml`: uses helper templates for consistent database/username configuration

### DIFF
--- a/charts/diode/Chart.yaml
+++ b/charts/diode/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: diode
 description: A Helm chart for Diode
 type: application
-version: 1.8.0
+version: 1.9.0
 appVersion: "1.5.0"
 home: https://github.com/netboxlabs/diode
 sources:
--- a/charts/diode/README.md
+++ b/charts/diode/README.md
@@ -2,7 +2,7 @@
 
 A Helm chart for Diode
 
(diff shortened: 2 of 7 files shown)


MicahParks (approved):

mfiedorowicz (approved):
Thanks @ecdobry

jajeffries (approved):

PR #406, opened 2025-09-27
fix: reconciler - remove hardcoded sslmode config

anshulkamath:
previously, ssl mode was not configurable when creating a connection to the postgres database. while the lack of SSL is acceptable for local deployments (i.e., containerized postgres in-cluster), we hit security restrictions when trying to connect to external databases (like RDS). this commit allows the operator to configure the ssl mode to their choosing.


mfiedorowicz:
@anshulkamath Thanks for contribution, could you sign the CLA, please?

mfiedorowicz (approved):
Thanks @anshulkamath 👍

PR #472, opened 2026-01-26
feat(reconciler): integrate graph DB with feature flag

mfiedorowicz:
⚠️ Experimental ⚠️

Add optional graph database integration to the reconciler service:

- Add ENABLE_GRAPH_DB feature flag (default: false)
- Add ENTITY_MATCHING_CONFIG_PATH for configurable matching rules
- Implement ProcessorOption functional options pattern for GraphBuilder
- Initialize GraphBuilder conditionally in main.go when feature enabled
- Extract entities to graph DB during ingestion (non-blocking)
- Add example entity_matching_config.yaml with default rules
- Update sample.env with new environment variables

When enabled, entities are stored in the graph database for relationship tracking. Graph extraction failures are logged but don't block main ingestion processing.

### DIFF
--- a/diode-server/reconciler/ingestion_processor.go
+++ b/diode-server/reconciler/ingestion_processor.go
@@ -69,6 +69,7 @@ type IngestionProcessor struct {
 	metrics            Metrics
 	cancel             context.CancelFunc
 	mx                 sync.Mutex
+	graphBuilder       *GraphBuilder // nil when ENABLE_GRAPH_DB is false
 }
 
 // IngestionLogToProcess represents an ingestion log to process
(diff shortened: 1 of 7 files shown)


diode-server/reconciler/ingestion_processor.go
```diff
@@ -436,6 +454,17 @@ func (p *IngestionProcessor) CreateIngestionLogs(ctx context.Context, ingestReq
 
 		p.metrics.RecordIngestionLogCreate(ctx, true)
 
+		// Extract graph data if graph DB is enabled (non-blocking, errors logged but not fatal)
+		if p.graphBuilder != nil {
```
jajeffries, line 458: nitpick/non blocking suggestion - setting a null implementation in main.go would mean we wouldn't need this check

jajeffries (approved):
Overall looks good. Happy for it to go in without that suggestion.

diode-server/reconciler/ingestion_processor.go
mfiedorowicz: Would require adding new abstraction/interface, hence going with YAGNI here as is

PR #478, opened 2026-02-04
feat: add optional pprof listeners to servers

paulstuart:
If the service pprof environment variable is set with a valid listening address:port it will provide http pprof endpoints for analyzing live servers.

In service of OBS-1991

### DIFF
--- a/diode-server/cmd/auth/main.go
+++ b/diode-server/cmd/auth/main.go
@@ -21,9 +21,6 @@ const (
 
 	// used by open telemetry metrics
 	telemetryServiceName = "netboxlabs/diode/auth"
-
-	// PProfEnvVar is the environment variable for the optional pprof server address
-	PProfEnvVar = "DIODE_AUTH_PPROF_ADDR"
 )
 
 func main() {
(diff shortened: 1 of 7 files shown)


diode-server/cmd/auth/main.go
```diff
@@ -20,6 +21,9 @@ const (
 
 	// used by open telemetry metrics
 	telemetryServiceName = "netboxlabs/diode/auth"
+
+	// PProfEnvVar is the environment variable for the optional pprof server address
+	PProfEnvVar = "DIODE_AUTH_PPROF_ADDR"
```
mfiedorowicz: Can we extract these (across all services in this PRs) into service corresponding configs and start pprof listener after config is processed, please?

paulstuart: Done. I did set a default to localhost so it should be on by default, if that's not acceptable I'll remove the default and have it only on when explicitly enabled by the config.

jajeffries (approved):

diode-server/pprof/pprof.go
```diff
@@ -0,0 +1,69 @@
...
+//   - /debug/pprof/
+//   - /debug/pprof/cmdline
+//   - /debug/pprof/profile
+//   - /debug/pprof/symbol
+//   - /debug/pprof/trace
+//
+// NOTE: it is a blocking call; run it in a separate goroutine.
+func Listen(ctx context.Context, logger *slog.Logger, addr string) {
+	if addr == "" {
+		logger.Error("pprof address cannot be empty")
```
mfiedorowicz, line 25: It will just log and continue making server addr empty string, should we fail here?

mfiedorowicz: unless we let it fail on L67

paulstuart: Good catch. It had a return before but got edited out.

PR #481, opened 2026-02-05
fix: remove defaults and allow distinct env names for each pprof env var

paulstuart:
Refinement for OBS-1991


jajeffries (approved):
lgtm

PR #486, opened 2026-02-09
refactor: graph package

mfiedorowicz:
- Extract entity graph domain (types, repository interface, builder, mocks) into a dedicated `graph/` package, decoupling it from `reconciler/` for direct use by future gRPC handlers
- Add postgres adapter (`dbstore/postgres/graph_repository.go`) implementing `graph.Repository`
- Introduce narrow `NodeFinder` interface for entity matcher, replacing direct `graph.Repository` dependency
- Rename `ExtractGraph` → `UpsertEntity` and return `*Node` to align with the `CreateEntity` RPC it will back
- Scaffold `graph.Reader` for read operations (`ListEntities`), separating reads from the write-focused `Builder`


paulstuart (approved):
LGTM

PR #489, opened 2026-02-11
fix: centralize Redis client options to prevent missing ACL username

RangerRick:
## What this does

Centralizes Redis client option construction so the ACL username is never
accidentally omitted. Previously each entrypoint built `redis.Options` inline;
when the username was left out, go-redis authenticated as the `default` Redis
user instead of the configured ACL user (causing `WRONGPASS` when the default
user is disabled).

- Adds a `NewRedisOptions` helper in `diode-server/tls/` that builds
  `redis.Options` with consistent TLS and ACL username/password handling.
- Adopts the helper at all three production call sites: `cmd/ingester`,
  `cmd/reconciler` (both the main and stream clients), and `reconciler/server.go`.

## Pre-PR review

Ran the analyzer suite (code review, semgrep, security audit, silent-failure,
variant hunt, simplifier). No P0/P1 findings — the core change is
behavior-preserving and is a net security improvement over the inline code.
All fixed items below are P2 quality cleanups.

### Fixed (P2)

- **`NewRedisOptions` signature** — replaced 6 positional params (two adjacent
  same-typed string pairs `host,port` / `username,password` were a
  transposition footgun, and 6 args exceeded the ≤5 guideline) with a
  `RedisParams` struct. Flagged independently by the security audit and the
  simplifier.
- **Redundant username guard** — `Username` is now assigned unconditionally in
  the struct literal; an empty value is identical to leaving it unset, and this
  better matches the helper's "never omit the username" intent.
- **Reconciler double TLS build** — the reconciler's two clients differ only by
  DB, so the options are built once and copied (overriding DB) instead of
  calling `NewRedisOptions`/`ToTLSConfig` twice and re-reading the CA file.
- **Error wrapping** — `reconciler/server.go` now wraps the options error with
  `%w` (was `%v`) so callers can inspect the underlying TLS error.
- **Test cleanup** — removed a duplicate table row and two standalone tests
  already covered by the `NewRedisOptions` table; intent comments preserved.

### Deferred

- 15 test-only `redis.Options{}` construction sites (in `ingestion_processor_test.go`,
  `component_test.go`, `component_internal_test.go`) were not migrated to the
  helper. They use miniredis with no ACL/TLS, so the default-user bug cannot
  manifest there. Left as-is by decision; not tracked separately.

### Reviewed and dismissed

- semgrep reported 7 warnings, all in unmodified files (TLS min-version in
  `netboxdiodeplugin/client.go`, unsafe usage in generated protobuf, plaintext
  HTTP in `telemetry/prometheus.go`) — pre-existing, outside this diff's scope.

Verified locally: `go build ./...`, `go vet`, `gofmt`, and `go test ./tls/ ./reconciler/`
all pass. (`golangci-lint` couldn't run locally due to a Go toolchain version
mismatch in the installed binary; CI runs the correct version.)

### DIFF
--- a/diode-server/tls/redis.go
+++ b/diode-server/tls/redis.go
@@ -6,29 +6,35 @@ import (
 	"github.com/redis/go-redis/v9"
 )
 
+// RedisParams holds the connection parameters for building redis.Options.
+type RedisParams struct {
+	Host     string
+	Port     string
+	Username string
+	Password string
+	DB       int
+	TLS      *Config
+}
+
 // NewRedisOptions builds redis.Options with consistent TLS and ACL username
 // handling. All Redis client creation should use this helper to ensure the
 // username is never accidentally omitted (which would cause go-redis to
 // authenticate as the "default" Redis user instead of the intended ACL user).
-func NewRedisOptions(host, port, username, password string, db int, tlsCfg *Config) (redis.Options, error) {
-	if tlsCfg == nil {
+func NewRedisOptions(p RedisParams) (redis.Options, error) {
+	if p.TLS == nil {
 		return redis.Options{}, fmt.Errorf("TLS config must not be nil (use &tls.Config{} for no TLS)")
 	}
 
-	goTLS, err := tlsCfg.ToTLSConfig()
+	goTLS, err := p.TLS.ToTLSConfig()
 	if err != nil {
 		return redis.Options{}, fmt.Errorf("failed to create TLS config for Redis: %w", err)
 	}
 
-	opts := redis.Options{
-		Addr:      fmt.Sprintf("%s:%s", host, port),
-		Password:  password,
-		DB:        db,
+	return redis.Options{
+		Addr:      fmt.Sprintf("%s:%s", p.Host, p.Port),
+		Username:  p.Username,
+		Password:  p.Password,
+		DB:        p.DB,
 		TLSConfig: goTLS,
-	}
-	if username != "" {
-		opts.Username = username
-	}
-
-	return opts, nil
+	}, nil
 }
--- a/diode-server/tls/redis.go
+++ b/diode-server/tls/redis.go
@@ -1,0 +1,34 @@
+package tls
+
+import (
+	"fmt"
+
+	"github.com/redis/go-redis/v9"
+)
+
+// NewRedisOptions builds redis.Options with consistent TLS and ACL username
+// handling. All Redis client creation should use this helper to ensure the
+// username is never accidentally omitted (which would cause go-redis to
+// authenticate as the "default" Redis user instead of the intended ACL user).
+func NewRedisOptions(host, port, username, password string, db int, tlsCfg *Config) (redis.Options, error) {
+	if tlsCfg == nil {
+		return redis.Options{}, fmt.Errorf("TLS config must not be nil (use &tls.Config{} for no TLS)")
+	}
+
+	goTLS, err := tlsCfg.ToTLSConfig()
+	if err != nil {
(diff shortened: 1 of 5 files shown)


RangerRick:
not sure why the go test stuff is failing, it's working for me

jajeffries (approved):
Overall looks good

diode-server/tls/redis.go
```diff
@@ -0,0 +1,34 @@
...
+
+	"github.com/redis/go-redis/v9"
+)
+
+// NewRedisOptions builds redis.Options with consistent TLS and ACL username
+// handling. All Redis client creation should use this helper to ensure the
+// username is never accidentally omitted (which would cause go-redis to
+// authenticate as the "default" Redis user instead of the intended ACL user).
+func NewRedisOptions(host, port, username, password string, db int, tlsCfg *Config) (redis.Options, error) {
+	if tlsCfg == nil {
```
mfiedorowicz, line 14: **🔵 Low**

Returning `redis.Options` by value copies the struct (which is fairly large). Consider returning `*redis.Options` to avoid the copy and to be consistent with how go-redis typically consumes options.

```suggestion
func NewRedisOptions(host, port, username, password string, db int, tlsCfg *Config) (*redis.Options, error) {
```

mfiedorowicz (commented):
## Argus Code Review Summary

- 🔵 Low: **1**

> 🔍 Automated review by **Argus** on behalf of @mfiedorowicz

diode-server/tls/redis.go
mfiedorowicz, line 14: **🔵 Low**

Returning `redis.Options` by value copies the entire struct (which includes a TLS config pointer, among other fields). Returning a pointer `*redis.Options` is more idiomatic for option structs and avoids unnecessary copying, especially since callers will likely pass `&opts` to `redis.NewClient()`.

```suggestion
func NewRedisOptions(host, port, username, password string, db int, tlsCfg *Config) (*redis.Options, error) {
```

mfiedorowicz, line 14: **🔵 Low**

The `port` parameter is a string, which means callers could pass invalid values (empty string, non-numeric). Consider accepting `port` as an `int` or validating it, or alternatively accept a single `addr` string parameter to match `redis.Options.Addr`.

mfiedorowicz (commented):
## Argus Code Review Summary

- 🔵 Low: **2**

> 🔍 Automated review by **Argus** on behalf of @mfiedorowicz

diode-server/tls/redis.go
```diff
@@ -0,0 +1,34 @@
...
+	"fmt"
+
+	"github.com/redis/go-redis/v9"
+)
+
+// NewRedisOptions builds redis.Options with consistent TLS and ACL username
+// handling. All Redis client creation should use this helper to ensure the
+// username is never accidentally omitted (which would cause go-redis to
+// authenticate as the "default" Redis user instead of the intended ACL user).
+func NewRedisOptions(host, port, username, password string, db int, tlsCfg *Config) (redis.Options, error) {
```
mfiedorowicz: **🔵 Low**

Returning redis.Options by value copies the entire struct (which includes a TLSConfig pointer and other fields). Consider returning *redis.Options to avoid the copy and to match how go-redis APIs typically consume options.

```suggestion
func NewRedisOptions(host, port, username, password string, db int, tlsCfg *Config) (*redis.Options, error) {
```

mfiedorowicz (commented):
## Argus Code Review Summary

- 🔵 Low: **1**

> 🔍 Automated review by **Argus** on behalf of @mfiedorowicz

mfiedorowicz (approved):

PR #496, opened 2026-03-05
feat: make Redis password optional across all services

mfiedorowicz:
Only set RedisPassword on redis.Options when the value is non-empty, matching the existing RedisUsername behavior.


paulstuart (approved):
LGTM

PR #501, opened 2026-03-23
feat: bulk write ingestion logs to reduce DB round-trips

mfiedorowicz:
## Summary
- Batch ingestion log creation using PostgreSQL `COPY` protocol to reduce per-entity DB round-trips
- Add `BulkCreateIngestionLogs` and `BulkUpdateIngestionLogs` repository methods with SQLC `copyfrom` support
- Refactor ingestion processor to collect results and write them in bulk after processing all entities in a stream read


diode-server/reconciler/ingestion_processor.go
```diff
@@ -229,6 +231,16 @@ func (p *IngestionProcessor) consumeIngestionStream(ctx context.Context, redisSt
 		}
 		b.Reset()
 
+		// ACK all messages immediately to prevent other instances from reclaiming
```
jajeffries, line 234: Does this mean if the server crashed at any point after this (because of updates/deployments causing a restart for example) then we'd lose those messages?

mfiedorowicz: Yes, that's correct.

jajeffries: Do we need to do anything to handle that? As I think that's a change in behaviour vs the current implementation

jajeffries (approved):
Approved and we can continue the conversation about handling messages that are being processed mid deployment separately. Probably a wider discussion point.

diode-server/reconciler/ingestion_processor.go
mfiedorowicz: Correction: when these messages are read and ack'd here, they are not lost, they still exist in the redis stream cause ACK doesn't delete a message (it requires additional DEL operation)

mfiedorowicz: The problem with current implementation is that when we read a message and restart/crash the service without ACKing the message, the message gets stuck in PEL state and is never reprocessed allocating redis memory indefinitely

mfiedorowicz: Will address it in a separate PR before the next release.

PR #503, opened 2026-03-24
refactor: concurrent Redis stream consumer with heartbeat and PEL drain

mfiedorowicz:
## Summary
- Concurrent worker pool for stream message processing (configurable `STREAM_WORKER_COUNT`, default 4)
- Heartbeat goroutine XCLAIMs in-flight messages to prevent false reclaim by other instances
- Two-phase startup: drain own PEL before consuming new messages
- Per-message ACK after processing, XDel after XAck
- Non-fatal worker errors (left in PEL for reclaimer)
- Graceful shutdown: drains in-flight workers before exit
- `ReclaimMinIdle` increased from 1min to 2min


jajeffries (approved):
Nice! LGTM

PR #508, opened 2026-03-29
fix: Preserve request-level metadata

MicahParks:
The purpose of this pull request is to preserve incoming request-level metadata.


diode-server/graph/metadata.go
mfiedorowicz, line 1: Please add tests for this change

mfiedorowicz (approved):

PR #522, opened 2026-05-07
feat(ingester): opt-in soft-shed when Redis memory crosses high-watermark

MicahParks:
Opt-in admission check on the ingester. When `REDIS_MEMORY_HIGH_WATERMARK_PCT` is set, `Ingest` rejects with `codes.ResourceExhausted` once Redis `used_memory` reaches that percentage of `maxmemory`. Disabled by default (`pct=0`).

Lets diode back off before saturating a Redis instance it may share with other workloads. Has no effect unless an operator has configured `maxmemory` + `maxmemory-policy noeviction` at the Redis layer.

The check is global on `used_memory` rather than per-tenant or per-stream — Diode's total share against the Redis pool is what needs to stay under budget when Redis is shared.

## Config

| Var | Default | Effect |
|---|---|---|
| `REDIS_MEMORY_HIGH_WATERMARK_PCT` | `0` | `0` disables; `1..100` rejects when used ≥ pct of maxmemory |
| `REDIS_MEMORY_CHECK_INTERVAL` | `500ms` | `INFO memory` cadence; bounds gauge freshness |
| `REDIS_MEMORY_CHECK_TIMEOUT` | `250ms` | Per-call timeout on `INFO memory`; prevents a hung Redis from stalling concurrent Ingest |

## Behavior

- Watermark check runs after request validation, so format errors surface as themselves rather than being masked by `ResourceExhausted`.
- `XAdd` OOM errors are also mapped to `ResourceExhausted`, so SDK callers see a consistent code whether the watermark soft-shed or Redis itself rejected.
- On `INFO` error or timeout the previous reading is retained rather than zeroed — a struggling Redis should not suddenly admit a flood.

## Metrics

- `ingest_redis_rejections_total{reason="watermark"|"redis_oom"}` — counter
- `redis_memory_used_ratio_bps` — gauge (0..10000)

## Test plan

- [x] Unit: `parseMemory`, watermark decision table, OOM detection, retain-on-INFO-error, configured interval, interval and timeout fallbacks, validation-before-watermark ordering, end-to-end `Ingest` OOM path, recorder wrappers
- [ ] Manual: with `maxmemory` set on Redis, drive ingest until rejected; confirm `ResourceExhausted` and rejection metric increments


jajeffries (approved):

mfiedorowicz:
Quick coordination question on the Helm side:

The new `REDIS_MEMORY_HIGH_WATERMARK_PCT` / `REDIS_MEMORY_CHECK_INTERVAL` / `REDIS_MEMORY_CHECK_TIMEOUT` env vars aren't first-class in `charts/diode/templates/diode-ingester-configmap.yaml` - operators would have to use `diodeIngester.extraEnvs` in `values.yaml` to enable the feature on a Helm-deployed cluster.

Two options to make it reachable out of the box:

1. **Add them as first-class entries** in `diode-ingester-configmap.yaml`, gated on a values key (e.g. `diodeIngester.redisMemoryHighWatermarkPct`), defaulting to disabled so the chart matches the binary's default behaviour.
2. **Document the `extraEnvs` path** in the PR / chart README and leave first-class wiring to a follow-up.

Which would you rather we do?

MicahParks:
> Quick coordination question on the Helm side:
> 
> The new `REDIS_MEMORY_HIGH_WATERMARK_PCT` / `REDIS_MEMORY_CHECK_INTERVAL` / `REDIS_MEMORY_CHECK_TIMEOUT` env vars aren't first-class in `charts/diode/templates/diode-ingester-configmap.yaml` - operators would have to use `diodeIngester.extraEnvs` in `values.yaml` to enable the feature on a Helm-deployed cluster.
> 
> Two options to make it reachable out of the box:
> 
> 1. **Add them as first-class entries** in `diode-ingester-configmap.yaml`, gated on a values key (e.g. `diodeIngester.redisMemoryHighWatermarkPct`), defaulting to disabled so the chart matches the binary's default behaviour.
> 2. **Document the `extraEnvs` path** in the PR / chart README and leave first-class wiring to a follow-up.
> 
> Which would you rather we do?

This change is very niche, most users will not need to consider this high-throughput failure state. However, I think it should live first-class in `diode-ingester-configmap.yaml`. I'll make an update.

mfiedorowicz (approved):

PR #526, opened 2026-05-13
feat(reconciler): persistent worker pools + bulk operations for decoupled ingestion processing

mfiedorowicz:
## Summary

Rework the reconciler's ingestion path from per-message channel/worker dispatch into a Postgres-backed inbox model with persistent worker pools. The Redis consume loop now drains at Postgres COPY speed and no longer blocks on NetBox HTTP; plan and apply phases run in their own decoupled processors with independent concurrency.

## ⚠️ Breaking change

This release **requires diode-netbox-plugin >= 1.12.0** (which ships `/bulk-plan` and `/bulk-plan-apply`). The previous `BULK_OPERATIONS_ENABLED` opt-in flag, the single-entity `/generate-diff` and `/apply-change-set` HTTP code paths, and the per-item `differ.Diff` fallback have all been removed - bulk endpoints are now the only code path. Deploy the plugin first, then upgrade diode-server.

## What changes

**Consume loop is now thin.** `handleStreamMessage` decodes the batch, bulk-COPYs new `ingestion_logs` rows (state=QUEUED), XACK/XDELs, and returns. No worker channels, no NetBox calls, no waiting on apply.

**Two new long-running processors:**

- `IngestionLogProcessor` - polls QUEUED rows via `FOR UPDATE SKIP LOCKED`, dispatches to a persistent worker pool. Calls `Ops.BulkPlan` (formerly `BulkGenerateChangeSets`) at its own pace. Configurable concurrency, batch size, and Redis-stream-length backpressure threshold.
- `AutoApplyProcessor` - combined plan + apply via the new `/bulk-plan-apply` HTTP call when `AUTO_APPLY_CHANGESETS=true`. Halves NetBox round-trips per entity vs separate plan + apply.

**Bulk-only API surface.** The reconciler now exposes a single, consolidated set of operations: `Ops.BulkPlan` and `Ops.BulkPlanApply`. The single-entity wrappers (`Ops.GenerateChangeSet`, `Ops.ApplyChangeSet`), the `applier` package, `differ.Diff`, and the netbox client's `GenerateDiff` / `ApplyChangeSet` (single) / `BulkApply` HTTP methods + their request/response types are all deleted. The `NetBoxAPI` interface shrinks from 6 methods to 3 (`BulkPlan`, `BulkPlanApply`, `GetDefaultBranch`); `IngestionProcessorOps` from 8 to 6.

**Bulk DB writes:** `BulkCreateIngestionLogs` (sqlc `:copyfrom`) and `BulkPersistChangeSets` replace per-row INSERT loops in the hot paths.

**Background default-branch refresher** owns the NetBox HTTP call that previously sat in the consume hot path. Hot-path callers read a cache; a goroutine refreshes on a fixed interval and retains the last-known value on transient failure, so consume keeps draining if NetBox/auth is briefly unreachable. `Ops.Start(ctx)` is idempotent via `sync.Once`.

**Proto / state machine:** new `APPLYING` state for the apply-in-flight window.

**Cleanup:** drops the dead `GenerateChangeSetConcurrency` / `ApplyChangeSetConcurrency` config (left over from the removed in-line worker dispatch) and the `BULK_OPERATIONS_ENABLED` compatibility flag now that bulk endpoints are required.

## Notes for review

- 26 commits, broken up by logical step (bulk-COPY paths, persistent worker pool, apply split, bulk-plan-apply, default-branch decoupling, dead-code cleanup, flag removal, API consolidation, restored differ unit tests). Reviewing commit-by-commit may be easier than the full diff.
- New tests: `ingestion_log_processor_test.go`, expanded `ops_test.go`, and a focused `differ/deviation_name_test.go` (internal test package) that restores unit-test coverage of the deviation-name helpers after `differ.Diff` was removed. Existing `ingestion_processor` tests trimmed to reflect the slimmer consume loop.

## Test plan

- [x] `make lint` clean
- [x] `make test` passes (race detector on)
- [x] Manual: NetBox briefly stopped -> consume loop keeps draining; warning logged once per refresh interval; resumes cleanly on NetBox restart
- [x] Manual: end-to-end load test against local stack with bulk endpoints


jajeffries (approved):

samiura (approved):

diode-server/reconciler/auto_apply_processor.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+func (p *AutoApplyProcessor) pollWorker(ctx context.Context) {
+	for {
+		select {
+		case <-ctx.Done():
+			p.logger.Debug("auto-apply processor exiting poll loop on request")
+			return
+		default:
+		}
```
MicahParks, line 172: This check can be removed. If context is expired, it will catch in `p.backpressure(ctx)`

diode-server/reconciler/ingestion_log_processor.go
```diff
@@ -0,0 +1,179 @@
...
+
+// Stop stops the ingestion log processor.
+func (p *IngestionLogProcessor) Stop() error {
+	p.logger.Info("stopping component", "name", p.Name())
+	p.mx.Lock()
+	if p.cancel != nil {
+		p.cancel()
+		p.cancel = nil
+	}
+	p.mx.Unlock()
```
MicahParks, line 134: A [`context.CancelFunc` does not need a lock](https://pkg.go.dev/context#CancelFunc)

> A CancelFunc tells an operation to abandon its work. A CancelFunc does not wait for the work to stop. A CancelFunc may be called by multiple goroutines simultaneously. After the first call, subsequent calls to a CancelFunc do nothing.

diode-server/reconciler/ingestion_log_processor.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+func (p *IngestionLogProcessor) pollWorker(ctx context.Context) {
+	for {
+		select {
+		case <-ctx.Done():
+			p.logger.Debug("ingestion log processor exiting poll loop on request")
+			return
+		default:
+		}
```
MicahParks, line 190: Another unneeded `ctx` check.

MicahParks (approved):

diode-server/reconciler/ingestion_log_processor.go
MicahParks: Unless this `.cancel` field is being reassigned? I didn't check for that.

mfiedorowicz: Right that `CancelFunc` itself is concurrent-safe - the lock here isn't protecting the `cancel()` call, it's protecting the field access on `p.cancel`:

- `Start` writes `p.cancel = cancel`
- `Stop` reads `p.cancel`, nil-checks it, calls it, writes `p.cancel = nil`

Concurrent `Start` + `Stop` on different goroutines would data-race on `p.cancel` without the lock. The `p.cancel = nil` re-assignment in `Stop` is what makes it a race rather than a safe read - your follow-up was on the money.

This is also the established pattern for `Component` implementations in this file - `IngestionProcessor.Start/Stop` (the pre-existing one) uses the exact same `mx.Lock` + nil-check + reassign idiom. Both new processors here (`IngestionLogProcessor`, `AutoApplyProcessor`) were modelled on it for consistency.

Two ways to simplify if we want to:

1. **Drop the nil-reset + nil-check entirely**, lean on `CancelFunc` idempotency, and drop the lock. Cleaner but needs to happen across all three components, not just this one.
2. **Keep as-is** for now, defer (1) to a follow-up sweep.

I'd lean (2) within the scope of this PR (to keep the three processors in lockstep) and file a small follow-up to do (1) consistently across the file. Happy to do (1) here if you'd rather flatten the convention now.

MicahParks (approved):

PR #550, opened 2026-07-06
fix(differ): return structured errors with plugin error details

mfiedorowicz:
Per-entity plan/apply failures from `/bulk-plan` and `/bulk-plan-apply` were flattened into `fmt.Errorf` strings, so the plugin's error list was lost as structure and the ingestion log message leaked internal jargon ("bulk plan error for entity 123: [...]").

`ConvertBulkPlanResult` and `ConvertBulkPlanApplyResult` now return `changeset.Error` with:
- a user-facing message: "failed to determine deviation" (plan) / "failed to apply deviation to NetBox" (apply)
- the matching error code (`ERR_OPS_GENERATE_DIFF` / `ERR_OPS_APPLY_CHANGE_SET`)
- the plugin's raw error list in `details`

`handleChangeSetError` passes `*changeset.Error` through untouched, so the ingestion log error column now stores `{message, code, details}` with the causes intact instead of everything squashed into the message. Entity attribution isn't lost — the error is persisted on its ingestion log and callers log the ID.

### DIFF
--- a/diode-server/reconciler/differ/differ.go
+++ b/diode-server/reconciler/differ/differ.go
@@ -6,6 +6,7 @@ import (
 	"github.com/google/uuid"
 	"google.golang.org/protobuf/proto"
 
+	diodeErrors "github.com/netboxlabs/diode/diode-server/errors"
 	"github.com/netboxlabs/diode/diode-server/gen/diode/v1/diodepb"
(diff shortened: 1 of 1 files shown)


jajeffries (approved):
LGTM

PR #551, opened 2026-07-06
feat(reconciler): requeue deduplicated entities for re-plan and apply

mfiedorowicz:
Deduplicated entities (same entity hash as a prior ingestion log) only incremented a duplicate counter — the prior log never re-entered the plan/apply pipeline, so NetBox state drift (entity modified or deleted after apply) went undetected until the entity data itself changed.

## Change

On duplicate detection, the new `BulkMarkDuplicates` query (replacing `IncrementDuplicateCount`/`BulkIncrementDuplicateCounts`) increments duplicate bookkeeping and atomically requeues the prior log back to `QUEUED` when its NetBox state may have drifted, recording the original state in the new `requeued_from_state` column:

| Prior state | Action |
|---|---|
| APPLIED, FAILED, NO_CHANGES | requeued → re-planned (and re-applied under auto-apply); `error` cleared |
| QUEUED, OPEN, APPLYING | untouched (already pending or in flight) |
| IGNORED | untouched (user opt-out) |

The existing pollers pick requeued rows up unchanged. The state guard is evaluated on the locked row inside the UPDATE, so a row claimed concurrently by a poller is correctly skipped; row locks are taken in id order to stay deadlock-safe with the claim queries and the state-count triggers.

Also adds an `ingestion_log_requeue_total` counter and a `Requeued` flag on `CreateIngestionLogResult`.

## Deviations are immutable: new deviation on drift

An applied deviation is a historical record of what diode did to NetBox — re-planning must not rewrite it. A log requeued from `APPLIED` therefore forks at persist time:

- **drift found** → a new ingestion log (clone of the prior via `CloneIngestionLogForDrift`: same entity/hash/source fields, fresh external ID and ingestion timestamp) is created carrying the drift change set — `OPEN` in the plan-only flow, `APPLIED`/`FAILED` in plan-apply — and the prior is restored to `APPLIED`. All in one transaction (`BulkCreateDriftDeviations`). Dedup lookups pick the newest log per entity hash, so subsequent duplicates track the new deviation.
- **no drift** → the prior is restored to `APPLIED` (not `NO_CHANGES` — the applied fact is preserved); nothing persisted.
- **plan failure** → the prior is restored to `APPLIED`; the error goes to Sentry/logs only and the re-check retries on the next observation. No `FAILED` placeholder overwrites the applied record.

Requeues from `FAILED`/`NO_CHANGES` keep re-planning in place — those logs have no applied history to protect. `requeued_from_state` is cleared on every final state write. New deviation rows scale with real drift, not observation frequency, so the dedup collapse is preserved.

Includes one migration: `ALTER TABLE ingestion_logs ADD COLUMN requeued_from_state INTEGER`.

## Verification

- Unit tests: requeue plumbing (single + bulk paths, error propagation), per-state duplicate handling, drift fork in both `BulkPlan` and `BulkPlanApply` (drift → new deviation; clean → restore APPLIED; plan failure → restore; apply failure → new FAILED deviation annotated with the error)
- SQL validated against Postgres 16 with all migrations applied (no DB integration tests exist in this repo): per-state requeue behavior with `requeued_from_state` recording, a concurrent-claim race (requeue blocks on the claim's row lock, then correctly skips the now-OPEN row), the full drift lifecycle (requeue → clone → restore) including `v_deviations` output, dedup picking the newest log per hash, and state-count trigger consistency

## Note

Permanently-failing entities retry on every observation cycle (FAILED → QUEUED → FAILED), each producing a Sentry capture — the new counter gives visibility if this gets noisy. A throttle was considered and deliberately left out for now.


jajeffries (approved):

jajeffries (approved):

diode-server/dbstore/postgres/queries/ingestion_logs.sql
```diff
@@ -91,16 +86,57 @@ VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14);
 -- name: FindIngestionLogIDsByExternalIDs :many
 SELECT id, external_id FROM ingestion_logs WHERE external_id = ANY(@external_ids::text[]);
 
--- name: BulkIncrementDuplicateCounts :exec
-UPDATE ingestion_logs
-SET duplicate_count = duplicate_count + 1,
-    last_seen = CURRENT_TIMESTAMP
```
davidlanouette, line 97: Did you intend to loose the update to `last_seen`?

davidlanouette: UG.  Ignore.  I can't read...

diode-server/dbstore/postgres/queries/ingestion_logs.sql
```diff
@@ -58,12 +59,6 @@ WHERE il.entity_hash = sqlc.arg('entity_hash')
 ORDER BY il.created_at DESC
 LIMIT 1;
 
--- name: IncrementDuplicateCount :exec
```
davidlanouette, line 61: Q: Do these get automatically deleted when you regenerate the sql layer?

mfiedorowicz: after sqlc generate deleted/modified queries get reflected in the Go code with these queries accordingly

PR #570, opened 2026-07-30
feat(auth): cache client credentials tokens in POST /token

mfiedorowicz:
## What

`POST /token` is a 1:1 proxy to the oauth2 server, so caller token rate is upstream token rate. Adds an optional per-credential cache with single-flight. Disabled by default.

## Safety

The presented secret is part of every cache key, so a valid `client_id` with the wrong secret derives a different key and always goes upstream. Only `client_credentials` and only `200` responses are cached. `expires_in` is recomputed on every hit. Reuse is bounded by `min(OAUTH2_TOKEN_CACHE_MAX_TTL, 80% of token lifetime)`.

Deployments running extra checks at issuance can install a `TokenIssuanceGate`, consulted on every hit. A denial or gate error falls through to the upstream rather than serving the entry, so failing closed costs a cache hit and not availability.

## Config

| var | default |
|---|---|
| `OAUTH2_TOKEN_CACHE_ENABLED` | `false` |
| `OAUTH2_TOKEN_CACHE_MAX_ENTRIES` | `4096` |
| `OAUTH2_TOKEN_CACHE_MAX_TTL` | `15m` |
| `OAUTH2_TOKEN_CACHE_NEGATIVE_TTL` | `5s` |

Memory measured locally at roughly 7 KiB per entry, so the default bound is about 27 MiB when full, and flat past capacity.

## Tests

14 new tests. Four are the ones that must never regress: a wrong secret is never served from cache, a narrower scope is never served from a broader entry, non-`client_credentials` grants are never cached, and a gate denial never serves the entry.


diode-server/auth/tokencache.go
```diff
@@ -0,0 +1,354 @@
...
+		cred.grantType,
+		cred.clientID,
+		cred.clientSecret,
+		strings.Join(cred.scopes, " "),
+		strings.Join(cred.audiences, " "),
+	} {
+		_, _ = fmt.Fprintf(mac, "%d:%s", len(field), field)
+	}
+	return base64.RawURLEncoding.EncodeToString(mac.Sum(nil))
+}
```
davidlanouette, line 145: NIT: Consider making this a function on the `tokenCredentials` struct, not on `tokenCache`.  It's a minor change, but I think it will encapsulate the logic better and make it cleaner to use.

Ex:

```suggestion
func (c *tokenCredentials) toCryptoKey(keyMaterial []byte) string {
	// Only calculate the key if it hasn't been already (since it's expensive)
	if c.key == "" {
		mac := hmac.New(sha256.New, keyMaterial)
		for _, field := range []string{
			c.grantType,
			c.clientID,
			c.clientSecret,
			strings.Join(c.scopes, " "),
			strings.Join(c.audiences, " "),
		} {
			_, _ = fmt.Fprintf(mac, "%d:%s", len(field), field)
		}
		c.key = base64.RawURLEncoding.EncodeToString(mac.Sum(nil))
	}
	return c.key
}
```

Then, you can use `key := cred.toCryptoKey(s.keyMaterial)` instead of `key := s.tokenCache.key(cred)`. You also wouldn't need to pass around key and creds, since the key is derivable from the creds.

diode-server/auth/tokencache.go
```diff
@@ -0,0 +1,354 @@
...
+	for _, field := range []string{
+		cred.grantType,
+		cred.clientID,
+		cred.clientSecret,
+		strings.Join(cred.scopes, " "),
+		strings.Join(cred.audiences, " "),
+	} {
+		_, _ = fmt.Fprintf(mac, "%d:%s", len(field), field)
+	}
+	return base64.RawURLEncoding.EncodeToString(mac.Sum(nil))
```
davidlanouette, line 144: Does the `Sum` need to be encoded?  Could you just have the key type on the cache be a `[]byte`?

Is there anywhere you are using this as a actual string - vs just the key to an element in the cache?

diode-server/auth/tokencache.go
```diff
@@ -0,0 +1,354 @@
...
+}
+
+// get returns a live entry for key. Entries past their serving window are removed and
+// reported as a miss.
+func (c *tokenCache) get(key string, now time.Time) (tokenCacheEntry, bool) {
+	entry, ok := c.entries.Get(key)
+	if !ok {
+		return tokenCacheEntry{}, false
+	}
+	if !now.Before(entry.servableUntil) {
```
davidlanouette, line 156: NIT: add a comment here saying why you are doing this.


```suggestion
	// Check if the entry is already expired.  If so, remove it from the cache and return and empty result.
	if !now.Before(entry.servableUntil) {
```

diode-server/auth/tokencache.go
```diff
@@ -0,0 +1,354 @@
...
+	})
+	return true
+}
+
+// putNegative caches an upstream client rejection.
+//
+// This is safe only because the presented secret is part of the key: a caller that later
+// presents the correct secret derives a different key and cannot be locked out by a
+// cached rejection.
+func (c *tokenCache) putNegative(key string, statusCode int, contentType string, body []byte, now time.Time) {
```
davidlanouette, line 198: Q: Are there any cases where this would be called, but there is an entry in the "normal" cache?

diode-server/auth/tokencache.go
```diff
@@ -0,0 +1,354 @@
...
+		scopes:       sortedFields(form["scope"]),
+		audiences:    sortedFields(form["audience"]),
+	}, true
+}
+
+// basicAuthCredentials extracts client_secret_basic credentials.
+//
+// RFC 6749 section 2.3.1 requires both halves to be form-urlencoded before they are
+// base64 encoded, which http.Request.BasicAuth does not undo.
+func basicAuthCredentials(header http.Header) (string, string, bool) {
```
davidlanouette: This doesn't appear to be used.

davidlanouette (approved):
Some questions and some suggestions.

But, I wouldn't complain too much about merging it as-is.

diode-server/auth/tokencache.go
mfiedorowicz: Careful with the suggestion as written: the guard is inverted. `if c.key != ""` only computes when the key is already set, so the first call returns `""` and every credential tuple collapses to the same empty key, sharing one cache entry regardless of credentials. It would want `== ""`.

On the memoisation itself I would skip it: `tokenCredentials` is constructed per request and never outlives it, so there is no second call to amortise, and it would need a pointer receiver plus mutable state on what is currently a value. HMAC-SHA256 over five short fields is sub-microsecond.

The encapsulation point is fair, but `keyMaterial` lives on the cache, so a method on `tokenCredentials` would take it as a parameter and just relocate the coupling. Happy to move it if you would still prefer that. This function also shrank a fair bit in 47f7cd9, see the thread on `basicAuthCredentials`.

diode-server/auth/tokencache.go
mfiedorowicz: A string is required regardless: `singleflight.Group.Do` takes a `string`, and the same key drives both the LRU and the single-flight group. `[32]byte` would satisfy the LRU alone but not that.

The base64 step is the genuinely optional part, since a Go string can hold arbitrary bytes. I kept it so the key stays printable if it ever ends up in a log line or a metric label. Happy to switch to raw bytes if you would rather not pay the encode.

diode-server/auth/tokencache.go
mfiedorowicz: Added in 47f7cd9, thanks.

diode-server/auth/tokencache.go
mfiedorowicz: No, and it is worth stating why, because `Add` would overwrite if it could happen. Every path that reaches the upstream has already dropped that key first: `get` removes on expiry, a gate denial removes before falling through, and an unusable response body removes too. So a cached rejection cannot replace a live token.

Added a comment in 47f7cd9 recording the invariant, since it currently holds by construction rather than by anything enforcing it.

diode-server/auth/tokencache.go
mfiedorowicz: It was used at `:247` and covered by the basic-auth case in `TestCacheableTokenRequest`, so it was not dead code. But you were pointing at the right lines for a different reason: `manager_hydra.go:45` pins every client this service creates to `client_secret_post`, and the authorization server rejects any other method, so basic credentials could only ever produce a 401. I verified that against a running stack.

So in 47f7cd9 I deleted the function and replaced the whole extraction path with a refusal to cache when an `Authorization` header is present. Net 27 lines lighter.

Keeping the detection still matters even though the extraction does not: dropping it entirely would mean keying on the form credentials while the upstream may authenticate the header, which could cache one client token under another client key. Confirmed against the running stack that the upstream does pick one when both are present.

diode-server/auth/tokencache.go
davidlanouette: Good catch (GH comments aren't a great IDE.)  I've fixed it.

> On the memoisation itself I would skip it:

Fair enough.  It would make it easier to not need to pass around the key too.  You could freely call `toCryptoKey` with no perf penalty.  But, that's a fairly minor thing.

davidlanouette (approved):
Looks great!

PR #579, opened 2026-08-27
fix(security): base auth on distroless/static, drop hydra binary

leoparente:
Follows #578, which moved ingester and reconciler to `distroless/static`. Auth was the last image on a package-managed base.

## Problem

Two separate CVE streams, both blocking or allowlisted:

**1. alpine openssl.** `alpine:3.22` ships `3.5.7-r0`; `3.5.8-r0` is available in v3.22/main and fixes CVE-2026-14456 (HIGH) plus 9 MEDIUM/LOW across `libssl3` and `libcrypto3`.

The Dockerfile *already* ran `apk upgrade --no-cache`. It wasn't taking effect: the scan workflow's `cache-from: type=gha` restores that `RUN` layer rather than re-executing it, so the upgrade had silently frozen at whatever was current when the layer was first built. Bumping the base tag would have re-frozen at the next advisory.

**2. The hydra binary.** The image copied `/usr/bin/hydra` from `oryd/hydra:v26.2.0`, carrying 45 findings — 2 CRITICAL (`jackc/pgx`), 43 HIGH (`x/crypto`, `otel`, `go-jose`) — which the workflow allowlisted via `isUpstreamHydra` because they need an upstream hydra release.

Both existed for one reason: `bootstrap-clients.sh` needed a shell, `jq`, and the hydra CLI.

## Change

**Port the bootstrap to Go.** New `authmanager bootstrap-clients` subcommand reads the same `client-credentials.json` and drives the same Hydra Admin API through the existing `auth.ClientManager`. Configuration is unchanged by construction — `HydraClientManager.CreateClient` already sets the identical `client_credentials` grant, `token` response type and `client_secret_post` auth method the script passed to `hydra create oauth2-client`. Log wording is preserved (`INFO: client X created` / `already exists`).

Two improvements over the script: it retries the admin API until it answers instead of `sleep 3`, and it treats a create `409` as "already exists" so concurrent runs are safe.

**Then the OS becomes unnecessary** — `gcr.io/distroless/static-debian13:nonroot`, matching ingester and reconciler. No shell, no jq, no hydra binary, no shared libraries. Image contents are now exactly `usr/local/bin/diode-server` and `usr/local/bin/authmanager`.

**Allowlist removed.** No hydra binary remains in any image, so `isUpstreamHydra` is deleted from the scan workflow and HIGH/CRITICAL block unconditionally again.

## ⚠️ Deployment coupling — needs a follow-up

The bootstrap job command changes to `["/usr/local/bin/authmanager", "bootstrap-clients"]` and reads `OAUTH2_ADMIN_SERVER_URL` instead of `HYDRA_ADMIN_URL`. Compose is updated in this PR.

`charts/diode/values.yaml` still pins `diodeAuthBootstrap.image.tag: 1.12.0`, which predates this subcommand. **That tag must be bumped to a release containing it before the chart is released.** The chart only publishes from the `release` branch and `ct` lints without installing, so nothing breaks on merge — but the bump can't happen in this PR because the version doesn't exist yet. I've left a comment on the `command:` line in the template.

## Verification

Fresh `--no-cache` build, scanned with CI's exact trivy invocation:

```
diode-auth:scan (debian 13.6)   debian    0 vulnerabilities
usr/local/bin/authmanager       gobinary  0 vulnerabilities
usr/local/bin/diode-server      gobinary  0 vulnerabilities
```

Image contents confirmed: zero `.so` files, no `/bin/sh`, no `jq`, no `usr/bin/hydra`.

Full dev stack brought up on the new image against a live NetBox:

- Bootstrap against an already-populated Hydra → `already exists` for all three clients, **exit 0**.
- Deleted `diode-ingest` via the new distroless `authmanager delete-client`, confirmed gone, re-ran bootstrap → `INFO: client diode-ingest created`, **exit 0**. Other two still skipped.
- The recreated client issued a token through nginx → `HTTP 200`, `scope: diode:ingest`.
- That token ingested a site over gRPC → `ingest OK`, reconciled to `state = 3 (APPLIED)` in `ingestion_logs`.

`go build ./...`, `go test ./...` and `golangci-lint run` all pass. `helm lint` clean; bootstrap job renders as expected. Unit tests cover create, skip-existing, 409-as-existing, lookup/create errors, credentials-file validation, and the readiness retry/give-up paths.


leoparente:
Superseded — going with a minimal fix instead. Keeping the hydra binary, the shell bootstrap and the helm chart untouched; only the stale openssl layer gets addressed.

PR #588, opened 2026-09-01
fix(scripts): strip trailing slashes from NETBOX_HOST

### LINKED ISSUE #563: quickstart.sh auto delete / in the end of NETBOX_HOST
### Diode version

v2.1.0

### Feature type

Change to existing functionality

### Proposed feature or enhancement

When I copy my netbox host, I get something like 'http://netbox.example.net/' (with / in the end)

I run `./quickstart.sh http://netbox.example.net/`

in the .env file, I get  `NETBOX_DIODE_PLUGIN_API_BASE_URL=http://netbox.example.net//api/plugins/diode` (with double slash //)

Then I get this error:
```diode-diode-reconciler-1      | {"time":"2026-07-22T15:13:55.30455738Z","level":"DEBUG","msg":"performing request","method":"GET","url":"
http://netbox.example.net//api/plugins/diode/default-branch/
"}
diode-diode-reconciler-1      | {"time":"2026-07-22T15:13:55.323344791Z","level":"DEBUG","msg":"get default branch","statusCode":404}
```

I think other people could make this mistake too, the quickstart.sh delete the trailing / automaticly is helpfull.

### Use case

./quickstart.sh http://netbox.example.net/

leoparente:
Fixes #563.

## Problem

`sample.env:37` holds the placeholder with the path already appended:

```
NETBOX_DIODE_PLUGIN_API_BASE_URL=<http://NETBOX_HOST>/api/plugins/diode
```

and the scripts substitute the argument verbatim:

```sh
sed "s|<http://NETBOX_HOST>|$NETBOX_HOST|g" "$ENV_FILE"
```

So a host copied from a browser address bar, which normally carries a trailing slash, yields `http://netbox.example.net//api/plugins/diode` and every plugin request 404s. The reconciler log points at NetBox rather than at the argument, so it doesn't read as a typo.

## Change

Strip trailing slashes before substitution, in both `quickstart.sh` and `configure-dev.sh` — they share the same argument handling.

```sh
while [[ "$NETBOX_HOST" == */ ]]; do
    NETBOX_HOST="${NETBOX_HOST%/}"
done
```

Placed **before** the existing empty check, so an argument of only slashes fails with `NETBOX_HOST is not provided` rather than writing a broken env file.

## Verification

Exercised the stripping plus the actual template substitution:

| input | resulting base URL |
|---|---|
| `http://netbox.example.net` | `http://netbox.example.net/api/plugins/diode` |
| `http://netbox.example.net/` | `http://netbox.example.net/api/plugins/diode` |
| `http://netbox.example.net///` | `http://netbox.example.net/api/plugins/diode` |
| `https://nb.example.com:8000/` | `https://nb.example.com:8000/api/plugins/diode` |
| `http://netbox.example.net/netbox/` | `http://netbox.example.net/netbox/api/plugins/diode` |
| `///` | empty, caught by the existing check |

The subpath case matters: NetBox served under `/netbox/` still resolves correctly, so this only removes the trailing separator rather than any path.

`bash -n` clean on both, `shellcheck -S warning` reports nothing on the changed files.

Thanks @tinyhare for the report, including the exact reproduction and the resulting log line.


leoparente:
@codex review

samiura (approved):

